PR #5, opened 2020-05-25
Alec fitdata

aabeaton:


nmrfit.py
```diff
@@ -0,0 +1,342 @@
...
+        return
+    def fitfunc_raw(self,p,x):
+        return p[0]/(p[1]+x)*x
+    def fitfunc_raw_symb(self,p,x):
+        try:
+            return p[0]/(p[1]+x)*x
+        except:
+            raise CustomError('types of p',p,map(type,p),'type of x',type(x),'x=',x)
+#}}}
+#}}}
```
jmfranck, line 342: we don't want to commit this to pyspecdata

pyspecdata/core.py
```diff
@@ -158,6 +159,8 @@ def mydiff(data,axis = -1):
 def normal_attrs(obj):
     myattrs = [x for x in dir(obj) if not ismethod(obj.__getattribute__(x))]
     myattrs = [x for x in myattrs if not x[0:2] == '__']
+    # next line filters out properties
+    myattrs = [x for x in myattrs if x not in ['C','angle','imag','real']]
```
jmfranck, line 163: I think the line currently in master is more up-to-date -- try `git diff master core`, and jump to line 163 in your current version to check it out.

I _think_ you should see there's a list defined that lists all the attributes... is so, that is the more up to date version

pyspecdata/core.py
```diff
@@ -2596,14 +2599,28 @@ def box_muller(length, return_complex=True):
...
+    Parameters
+    ----------
+    decimalplaces: int (optional, default 3)
+        number of decimal places
+    scientific: boolean (optional, default False)
+        use scientific notation
+    max_front: int (optional, default 3)
+        at most this many places in front of the decimal before switching
+        automatically to scientific notation.
+    """
```
jmfranck, line 2614: yes, this is nice

pyspecdata/core.py
```diff
@@ -6714,6 +6716,16 @@ def _init_datanode(self,datanode,**kwargs):
...
+    def __init__(self,pathstring,directory='.'):
+        self.pathstring = pathstring
+        #try:
+        self.h5file,self.datanode = h5nodebypath(pathstring,
+                check_only=True, directory=directory)
+        #except BaseException  as e:
+        #    raise IndexError("I can't find the node "+pathstring+explain_error(e))
+        print("about to call _init_datanode")
+        self._init_datanode(self.datanode)
+        atexit.register(self._cleanup)
```
jmfranck, line 6728: don't quite understand how this got moved around (though I probably did it?) .. I would move it up above init_datanode, so the diff is simpler

pyspecdata/core.py
```diff
@@ -6850,15 +6862,16 @@ def myfilter(x,center = 250e3,sigma = 100e3):
 
 #{{{ fitdata
 class fitdata(nddata):
+    r''' Inherits from an nddata and enables curve fitting through use of a sympy expression.
+    '''
```
jmfranck, line 6866: this is where I would add a little bit of documentaiton

pyspecdata/core.py
jmfranck: although, looking below it's possible this is the simpler version, since I notice we just re-use normal_attrs several times

pyspecdata/core.py
```diff
@@ -6887,132 +6900,55 @@ def __init__(self,*args,**kwargs):
...
-            #covarmatrix = array((J.T * G.T * G * J)**-1 * J.T * G.T * G * S * G.T * G * J * (J.T * G.T * G * J)**-1)
-            #try:
-            #    betapremult = (J.T * Omegainv * J)**-1 * J.T * Omegainv
-            #except:
-            #    print 'from sigma','\n\n',diag(sigma**2),'\n\n','from covarmatrix','\n\n',S,'\n\n'
-            #    raise RuntimeError('problem generating estimator (word?)')
-            #covarmatrix = array( betapremult * S * betapremult.T)
-        #print "shape of fprime",shape(fprime),"shape of fprime_prod",shape(fprime_prod),'sigma = ',sigma,'covarmat=',covarmatrix,'\n'
+    @property
+    def function_string(self):
```
jmfranck, line 6904: I would add a docstring here.. I would say the docstring isn't required for the setter, etc, but should be present here

pyspecdata/core.py
```diff
@@ -6887,132 +6900,55 @@ def __init__(self,*args,**kwargs):
         self.active_indices = None
         #}}}
         return
-    def parameter_derivatives(self,xvals,set = None,set_to = None):
-        r'return a matrix containing derivatives of the parameters, can set dict set, or keys set, vals set_to'
```
jmfranck, line 6891: don't we still want a `parameter_derivatives` function?  Even if a placeholder that returns an error for now?

pyspecdata/core.py
```diff
@@ -6887,132 +6900,55 @@ def __init__(self,*args,**kwargs):
...
-            except Exception as e:
-                raise ValueError(strm('Trying to set index',j,
-                    'shape(fprime)',shape(fprime),
-                    'shape(xvals)',shape(xvals))+explain_error(e))
-        return fprime
-    def parameter_gradient(self,p,x,y,sigma):
-        r'this gives the specific format wanted by leastsq'
-        # for now, I'm going to assume that it's not using sigma, though this could be wrong
-        # and I could need to scale everything by sigma in the same way as errfunc
-        return self.parameter_derivatives(x,set = self.symbol_list,set_to = p).T
```
jmfranck, line 6933: I would agree this is not required

pyspecdata/core.py
```diff
@@ -6887,132 +6900,55 @@ def __init__(self,*args,**kwargs):
...
-    def analytical_covariance(self):
-        covarmatrix = zeros([len(self._active_symbols())]*2)
```
jmfranck, line 6935: I would leave `analytical_covariance` as-is in the master branch -- maybe add a docstring saying it's not up to date, etc.

pyspecdata/core.py
```diff
@@ -7255,25 +7209,26 @@ def rename(self,previous,new):
...
-        if isinstance(set, dict):
-            set_to = list(set.values())
-            set = list(set.keys())
+        if isinstance(set_what, dict):
+            set_to = list(set_what.values())
+            set_what = list(set_what.keys())
         x = self.getaxis(self.fit_axis)
         if iscomplex(self.data.flatten()[0]):
-            if not silent: print(lsafen('Warning, taking only real part of fitting data!'))
+            if not silent: print((lsafen('Warning, taking only real part of fitting data!')))
```
jmfranck, line 7221: I would look for things in this function that refer to the variable `silent` and instead just do `logger.info(strm(....))` or `logger.debug(strm(...))`, as you think appropriate

pyspecdata/core.py
```diff
@@ -7255,25 +7209,26 @@ def rename(self,previous,new):
...
         y = real(self.data)
         sigma = self.get_error()
         if sigma is None:
             if not silent: print('{\\bf Warning:} You have no error associated with your plot, and I want to flag this for now\n\n')
             warnings.warn('You have no error associated with your plot, and I want to flag this for now',Warning)
             sigma = ones(shape(y))
-        p_ini = real(array(self.guess())) # need the numpy format to allow boolean mask
-        if set != None:
-            self.set_indices,self.set_to,self.active_mask = self.gen_indices(set,set_to)
+        p_ini = [1.0,1.0,1.0] # hard-coded for debug
```
jmfranck, line 7228: here, I would say that this is a fine default guess, but rather than hard-coding it like this, we should allow the user to set a dictionary that gives the initial guess variables.  The code actually won't work for anything that has  more than 3 parameters without doing this.

If this requires further discussion, I would set this to be `[1.0]*number_of_parameters` (where `number_of_parameters` is the appropriate variable for the number of parameters, and we can address this in a subsequent pull request

pyspecdata/core.py
```diff
@@ -7255,25 +7209,26 @@ def rename(self,previous,new):
...
+        #p_ini = real(array(self.guess())) # need the numpy format to allow boolean mask
```
jmfranck, line 7229: I would just delete lines that have been commented out

pyspecdata/core.py
```diff
@@ -7282,32 +7237,33 @@ def fit(self,set = None, set_to = None, force_analytical = False, silent = False
...
+        #        'type(sigma)',showtype(sigma),'shape(x):',shape(x),
+        #        'shape(y):',shape(y),'shape(sigma)',shape(sigma),
+        #        'p_ini',type(p_ini),p_ini))
+        #except ValueError as err:
+        #    raise ValueError(strm('leastsq failed with "',err,
+        #        '", maybe there is something wrong with the input:',
+        #        self))
+        #except Exception as e:
+        #    raise ValueError('leastsq failed; I don\'t know why'+explain_error(e))
+        ##}}}
```
jmfranck, line 7266: does it run if you uncomment these? if so I would leave this in

pyspecdata/core.py
```diff
@@ -7282,32 +7237,33 @@ def fit(self,set = None, set_to = None, force_analytical = False, silent = False
...
+        #        'type(x):',showtype(x),'type(y):',showtype(y),
+        #        'type(sigma)',showtype(sigma),'shape(x):',shape(x),
+        #        'shape(y):',shape(y),'shape(sigma)',shape(sigma),
+        #        'p_ini',type(p_ini),p_ini))
+        #except ValueError as err:
+        #    raise ValueError(strm('leastsq failed with "',err,
+        #        '", maybe there is something wrong with the input:',
+        #        self))
+        #except Exception as e:
+        #    raise ValueError('leastsq failed; I don\'t know why'+explain_error(e))
```
jmfranck, line 7265: here, I would remove `explain_error(e)` -- this is not required in Python 3, and actually will create confusion

(in Python 2,  when you caught and exception like this, it wouldn't tell you about the original error that caused it -- hence `explain_error`, which prints out info about the original error -- in Python 3, it does print the original error, so you get the same info twice, which is super confusing)

pyspecdata/core.py
```diff
@@ -7413,99 +7369,6 @@ def bootstrap(self,points,swap_out = exp(-1.0),seedval = 10347,minbounds = {},ma
...
-        which_starting_guess = 0
-        thisguess = self.starting_guesses[which_starting_guess]
-        numguesssteps = 20
-        #{{{ for some reason (not sure) adding a dimension to y
-        new_y_shape = list(y.shape)
-        new_y_shape.append(1)
-        y = y.reshape(tuple(new_y_shape))
-        #}}}
-        #{{{ evaluate f, fprime and residuals
-        guess_dict = dict(list(zip(self.symbol_list,list(thisguess))))
```
jmfranck, line 7436: this is the type of dictionary I was talking about above

pyspecdata/core.py
```diff
@@ -7413,99 +7369,6 @@ def bootstrap(self,points,swap_out = exp(-1.0),seedval = 10347,minbounds = {},ma
                         recordlist[runno][name] = thiscopy.output(name)
         print(r'\end{verbatim}')
         return recordlist # collect into a single recordlist array
-    def guess(self):
-        r'''provide the guess for our parameters; by default, based on pseudoinverse'''
```
jmfranck, line 7417: I would not delete this -- I actually think it could be quite useful for this dictionary of guesses that we are talking about.

pyspecdata/core.py
```diff
@@ -7282,32 +7237,33 @@ def fit(self,set = None, set_to = None, force_analytical = False, silent = False
             leastsq_kwargs.update({'Dfun':self.parameter_gradient})
```
jmfranck, line 7237: so, my understanding is that this doesn't work right now? (We are not generating Dfun correctly?)

I would recommend to make sure that you have your most up-to-date version of your attempt at Dfun using sympy diff saved in a branch that's different from `alec_fitdata` (typically, just check out the correct commit, and do `git checkout -b`, or even easier right-click on the correct commit in gitk and do "create branch"), and then please generate a new pull request using that branch.

nmrfit.py
aabeaton: agreed

aabeaton: we do not want any of this, correct?

pyspecdata/core.py
aabeaton: The difference between core here and core in master is that you have removed lines 162 and 163 here, but nothing was added in its place

pyspecdata/core.py
aabeaton: Glad you like it, but I do not think I added it

pyspecdata/core.py
aabeaton: Yes, this was your doing as I have not touched this (and diffed against core in master, it is exactly where you are saying to move it) -- so I have moved it there now

pyspecdata/core.py
aabeaton: Yes, agreed

pyspecdata/core.py
aabeaton: Ok, I have added this

pyspecdata/core.py
aabeaton: Yes not sure how this got deleted -- looking back at my notes, at one point this was giving me an error, but I do not think I resolved it by deleting it?

nmrfit.py
jmfranck: no -- none if it -- if we want this for reference, I have this file in my notebook

pyspecdata/core.py
aabeaton: Ok -- if we agreed to delete this, the above may have been deleted by mistake

pyspecdata/core.py
aabeaton: Ok, I just put this back in core on my branch

pyspecdata/core.py
jmfranck: actually, I don't see that difference now -- maybe pull down the most recent version of master?

pyspecdata/core.py
aabeaton: And have included a docstring saying not up to date

pyspecdata/core.py
jmfranck: yes, it looks like you don't have the most recent version of master.  PS, one of two ways of doing this -- 1) `git checkout master`, then `git pull`, then `git checkout alec_fitdata` or 2) `git fetch origin master:master` -- assuming you haven't made any changes to master locally, this will update it (note you can tab-complete the master:master, so it's not so much typing)

pyspecdata/core.py
aabeaton: Ok, here I have replaced the print statement with logger statements, as that is what I took this to mean

pyspecdata/core.py
jmfranck: I can't see any changes yet -- definitely feel free to make very incremental commits and push them up, because then I can go through these on the fly

pyspecdata/core.py
aabeaton: I agree this hard-coded approach is not the way to go moving forward. I like the number_of_parameters idea.

aabeaton: Alright, so this worked by just plugging in number of parameters as 3 -- now just need to find a way to get it to pull the number of parameters from what is provided by the user

aabeaton: OK, so I define a new variable self.number_of_parameters that pulls from functional_form when we are setting up the expression, and this appears to work (without hard-coding number_of_parameters as 3)

pyspecdata/core.py
aabeaton: Yes, did that while addressing the previous

pyspecdata/core.py
aabeaton: I uncommented these and it did run, so I am leaving them

pyspecdata/core.py
aabeaton: Ok, just removed explain_error

pyspecdata/core.py
aabeaton: yes, I know there was a discussion of a dictionary, I will look into this further

pyspecdata/core.py
aabeaton: Yes, Dfun does not work. I tried a few things quite a while ago and did not have success.

aabeaton: I will checkout a new branch then., it is likely stored in the git tree of this branch. I will check now.

pyspecdata/core.py
jmfranck: this sounds like a good solution

pyspecdata/core.py
jmfranck: I don't see a parameter derivatives function right now

pyspecdata/core.py
aabeaton: Ok - after going through the git tree, I found what was my stopping point for debugging Dfun and have checked that out as its own branch. Does this need to be pushed?

pyspecdata/core.py
jmfranck: I deal with this in a new pull request.

nmrfit.py
aabeaton: Alright - I just deleted it and pushed

pyspecdata/core.py
aabeaton: Cool this is done

pyspecdata/core.py
aabeaton: Yes - I am pushing them now as I do them

pyspecdata/core.py
jmfranck: yes, please push that, and then set it up as a pull request

pyspecdata/core.py
```diff
@@ -4174,10 +4184,6 @@ def argmin(self,*axes,**kwargs):
                     axis=thisindex)]
             self._pop_axis_info(thisindex)
         return self
-    def max(self):
-        return self.data.max()
-    def min(self):
-        return self.data.min()
```
jmfranck, line 4180: this seems to have been done in 5e5fb95a875672d.  It seems like 5e5fb95a875672d is rolling back changes that were going in the right direction.  I would revert 5e5fb95a875672d.

jmfranck: so actually, in db88b74, I'm trying to fix some of those changes, so first revert db88b74 then revert 5e5fb95

pyspecdata/core.py
aabeaton: Ok, it has been pushed + set up as pull request

aabeaton: To me there are 4 lines of code that are obvious for me to delete (which I am putting below) -- but these call things like parameter_gradient, so should that be deleted too?
`        if hasattr(self,'has_grad') and self.has_grad == True:
            leastsq_kwargs.update({'Dfun':self.parameter_gradient})
        if 'Dfun' in list(leastsq_kwargs.keys()):
            if not silent: print("yes, Dfun passed with arg",leastsq_kwargs['Dfun'])
`

pyspecdata/core.py
aabeaton: Does this need to be done in master? And by revert you mean to take the entirety of core.py and make it go back to its state in db88b74, then make it go back to its state in 5e5fb95?

pyspecdata/core.py
jmfranck: I think so.

Remember that you saved a branch where you are actually using `diff` to calculate the parameter gradients, and all this stuff (should?) still be there.

pyspecdata/core.py
jmfranck: I mean than on the alec branch, I would do:
`git revert db88b74` then `git revert 5e5fb95`
then do any debugging needed to get it to work properly.

I would just do this, but I'm worried that if we both did it at the same time, etc, etc.

jmfranck: to clarify, this rolls back the changes in 5e5fb95 (which itself appears to be some type of revert), since I'm claiming those are bad changes.  You need to do db88 first because I was manually correcting some of those changes, and when I tried this myself, not rolling back db88 first makes git think there is a conflict

pyspecdata/core.py
```diff
@@ -3306,10 +3321,7 @@ def set_error(self,*args):
         
         `set_error('axisname',error_for_axis)` or `set_error(error_for_data)`
 
-        `error_for_data` can be a scalar, in which case, **all** the data errors are set to `error_for_data`
-        .. todo::
-                several options below -- enumerate them in the documentation
-        '''
```
jmfranck, line 3312: I would leave the todo message (or, better, explain the ways it can be called)

pyspecdata/core.py
```diff
@@ -3321,8 +3333,6 @@ def set_error(self,*args):
             self.data_error = reshape(array(args[0]),shape(self.data))
         elif (len(args) is 2) and (isinstance(args[0], str)) and (isinstance(args[1], ndarray)):
             self.axis_coords_error[self.axn(args[0])] = args[1]
-        elif (len(args) is 2) and (isinstance(args[0], str)) and (isscalar(args[1])):
-            self.axis_coords_error[self.axn(args[0])] = args[1]*ones_like(self.getaxis(args[0]))
```
jmfranck, line 3325: this was added by me, please leave

pyspecdata/core.py
jmfranck: OK, sorry, I got confused, and thought you had already reverted the bad commit, but this was part of it

pyspecdata/core.py
jmfranck: OK, sorry, I got confused, and thought you had already reverted the bad commit, but this was part of it

pyspecdata/core.py
aabeaton: Yes, that is correct. so I will delete this bit of code and the associated.

pyspecdata/core.py
aabeaton: I see - that makes sense. I will begin this now.

PR #7, opened 2020-05-26
(NOT TESTED) -- suggested changes for alec_fitdata

jmfranck:


pyspecdata/core.py
```diff
@@ -6871,15 +6872,28 @@ def myfilter(x,center = 250e3,sigma = 100e3):
...
     ```
     import sympy as s
     s.init_printing()
     ```
-    you can then use `s.symbols(` to create symbols/variables that allow you to build the mathematical expression for your fitting function
-    """
+    you can then use `s.symbols(` to create symbols/variables that
+    allow you to build the mathematical expression for your fitting
+    function
+    '''
```
jmfranck, line 6896: this is a syntax fix.

note that I'm generating a conflict with the updates to the docstrings.

because these were long run-on lines, I can't easily tell what's going on here -- maybe just keep the suggested update?

pyspecdata/core.py
```diff
@@ -6918,20 +6932,25 @@ def __init__(self,*args,**kwargs):
         return
     @property
     def function_string(self):
-        r'''A property of the fitdata class which stores a string output of the functional form of the desired fit expression provided in func:`functional_form` in LaTeX format'''
+        r'''A property of the fitdata class which stores a string
+        output of the functional form of the desired fit expression
+        provided in func:`functional_form` in LaTeX format'''
```
jmfranck, line 6937: I would specify that this is read-only

pyspecdata/core.py
```diff
@@ -6918,20 +6932,25 @@ def __init__(self,*args,**kwargs):
...
         return r'$f(%s)='%(sympy.latex(self.fit_axis)) + retval + r'$'
     @function_string.setter
     def function_string(self):
         raise ValueError("You cannot set the string directly -- change the functional_form property instead!")
     @property
     def functional_form(self):
-        r'''A property of the fitdata class which is set by the user, takes as input a sympy expression of the desired fit expression'''
+        r'''A property of the fitdata class which is set by the user,
+        takes as input a sympy expression of the desired fit
+        expression'''
```
jmfranck, line 6947: I would give a basic example here, or in the fitdata docstring (if the latter, you can refer to it here as ``class:`fitdata```)

pyspecdata/core.py
```diff
@@ -6918,20 +6932,25 @@ def __init__(self,*args,**kwargs):
...
         print("Getting symbolic function")
         return self.symbolic_expr
     @functional_form.setter
     def functional_form(self,sym_expr):
-        r''' The functional form, given as a sympy expression, to which you would like to fit the data.'''
+        r''' The functional form, given as a sympy expression, to
+        which you would like to fit the data.'''
         assert issympy(sym_expr), "for now, the functional form must be a sympy expression!"
```
jmfranck, line 6954: I'm not positive, but I think you only need to give a docstring in the getter (the first definition of functional_form, above, which you've already done)

pyspecdata/core.py
```diff
@@ -7312,8 +7331,8 @@ def fit(self,set_what = None, set_to = None, force_analytical = False, silent =
             if not silent: print('{\\bf Warning:} You have no error associated with your plot, and I want to flag this for now\n\n')
             warnings.warn('You have no error associated with your plot, and I want to flag this for now',Warning)
             sigma = ones(shape(y))
-        if set_what != None:
-            p_ini = [1.0]*self.number_of_parameters
+        if set_what is not None:
+            p_ini = self.guess()
```
jmfranck, line 7335: seems more roundabout, but I'm doing this so that `guess()` can do other things, as well

pyspecdata/core.py
aabeaton: Are you saying to remove the docstring here that I wrote and keep yours? It is a simple syntax error as you point out -- I have combined the two, but I can remove mine if that is what you are saying.

pyspecdata/core.py
aabeaton: Ok this is done

pyspecdata/core.py
aabeaton: Ok I removed this here and added a little bit more in the getter docstring

pyspecdata/core.py
jmfranck: so what I would specifically suggest is:
1) highlight your docstring (shift-V) and do `gq` to get a decent wrapping.
2) compare the docstrings to make sure everything useful from both is included (just do `git diff alec_fitdata pyspecdata/core.py` if you're on this branch, or vice versa

pyspecdata/core.py
jmfranck: 👍

pyspecdata/core.py
aabeaton: I have already merged the branch, but I did precisely this and just commenting here about it for the sake of completeness.

PR #9, opened 2020-05-27
Logging

guinneak198:
Before we were not calling init_logging. The function now accepts a string, "info" or "debug" to specify what level we're logging. Example in log_test.py


pyspecdata/log_test.py
```diff
@@ -0,0 +1,11 @@
+from pyspecdata import *
+import logging
+#logging.basicConfig(stream=sys.stdout,level=logging.DEBUG)
+#logger.info("hello")i
+
+
+
+mylogger = init_logging("info")#print("hello")
+
+logger.info("Something, something, something, dark side...")
```
jmfranck, line 10: when I run, I get an error in my result
````
INFO:pyspecdata.core:Something, something, something, dark side...
--- Logging error ---
Traceback (most recent call last):
  File "C:\ProgramData\Anaconda3\lib\logging\__init__.py", line 1028, in emit
    stream.write(msg + self.terminator)
AttributeError: 'str' object has no attribute 'write'
Call stack:
  File "log_test.py", line 4, in <module>
    logger.info("Something, something, something, dark side...")
Message: 'Something, something, something, dark side...'
Arguments: ()
--> log_test.py(4):pyspecdata.core             <module> 2020-05-27 20:06:57,138
INFO: Something, something, something, dark side...
```

pyspecdata/general_functions.py
```diff
@@ -128,13 +128,38 @@ def check_ascending_axis(u,tolerance = 1e-7,additional_message = []):
     assert du > 0, thismsg
     return du
 
-def init_logging(level=logging.INFO, filename='pyspecdata.log'):
+def init_logging(levelString, filename='pyspecdata.log'):
```
jmfranck, line 131: First, AWESOME KUDOS for starting a pull request on pyspecdata!!

pyspecdata/log_test.py
jmfranck: to me, this says that when it expects a `stream` object, it's getting a string for some reason instead

pyspecdata/general_functions.py
```diff
@@ -128,38 +128,27 @@ def check_ascending_axis(u,tolerance = 1e-7,additional_message = []):
...
+def init_logging(level=logging.INFO, filename='pyspecdata.log'):
+    r"""Initialize logging on pyspecdata.log -- do NOT log if run from within a
+    notebook (it's fair to assume that you will run first before embedding)"""
+    if type(level) is str:
+        if level.lower() == 'info':
+            level=logging.INFO
+        elif level.lower() == 'debug':
+            level=logging.DEBUG
+        else:
+            raise ValueError("if you give me level as a string, give me 'info' or 'debug'")
```
jmfranck, line 140: the strategy is to interpret "level" here, and then just use it below.  A few less lines than what you ended up with, but does the same thing -- also what's below is less complicated

pyspecdata/general_functions.py
```diff
@@ -138,21 +138,23 @@ def init_logging(level=logging.INFO, filename='pyspecdata.log'):
...
     logger = logging.getLogger()
-    #fileHandler = logging.FileHandler(filename)
-    stdout_handler = logging.StreamHandler(sys.stdout)        
-    stdout_handler.setLevel(level)
-    #logger.setLevel(level)
-    formatter = logging.Formatter("--> %(filename)s(%(lineno)s):%(name)s %(funcName)20s %(asctime)20s\n%(levelname)s: %(message)s")
-    #fileHandler.setFormatter(logging.Formatter(FORMAT))
+    logger.setLevel(level) # even if I set the handler level, it won't
+    #                        print w/out this
+    file_handler = logging.FileHandler(log_filename, mode='a')
```
jmfranck, line 151: The problem mentioned above was that you want a FileHandler, which takes a file name, rather than a stream object (sys.stdout is a stream object)

examples/log_test.py
```diff
@@ -1,5 +1,4 @@
 from pyspecdata import *
-import logging
```
jmfranck: because I have it returning the logger object, which has info, debug, etc, methods, there is no longer a need to import logging.

this time, do feel free to resolve my comments as you read them, as long as you know what's going on, and post a comment once you've had a chance to review everything, and you can verify that this works for you.

pyspecdata/general_functions.py
guinneak198: I was running into an error where the obj had no attribute "lower" and upon looking into this lower I could not find much. Could you maybe help explain this?

pyspecdata/general_functions.py
guinneak198: yes! much more organized!

pyspecdata/general_functions.py
guinneak198: I THINK this section is saying that if the lowest level reads 'info' (as in logger.info()) then the level will be that of INFO or 20 (not quite sure on the numerical conception but working on it). If there is no string to state the lowest level then it won't declare a level

pyspecdata/general_functions.py
guinneak198: Okay so the logger = logging.getLogger() we NEED because we should NOT represent  loggers directly but through this module-level function (according to https://docs.python.org/3/library/logging.html).
the setLevel will set the threshold to level, so that if the logging message is less important than the level it will just be ignored. (same resource as above)
FileHandler will write the log in a file. But I believe to display it we need FileStreamer?

guinneak198: so I see the StreamHandler is addressed after this and this totally makes sense now

pyspecdata/general_functions.py
jmfranck: actually, you'll notice that in my edits, unrelated to what you were doing, I made the code more robust by checking for the case where `level` is set to a string.  `lower` is a method of the string class -- it converts everything to lowercase: This way, setting `level` to `"DEBUG"`, `"Debug"`, and `"debug"` has the same result

pyspecdata/general_functions.py
jmfranck: `getLogger()` gets the logger object that you can use to set the formatting, attach handlers, etc.

sounds like you are saying you get the difference between `StreamHandler` and `FileHandler` --> obviously, this is a one-off that you probably won't have to deal with again, so it's not about the specifics going on here --> the skill to really take away from the experience is how you look at the documentation/stackexchange/etc to figure out how to use those two functions on the fly.

pyspecdata/general_functions.py
guinneak198: ah okay this makes sense because I was actually wondering what the difference was between all caps and all lower case. But now my question is INFO and DEBUG are integers correct? so how can an integer have a method of lower?

pyspecdata/general_functions.py
guinneak198: yes much better

examples/log_test.py
guinneak198: makes sense

pyspecdata/log_test.py
guinneak198: So with the new changes that I just pushed, I get this error:
```
Traceback (most recent call last):
  File "log_test.py", line 2, in <module>
    mylogger = init_logging("info")
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\general_functions.py", line 134, in init_logging
    if type(level) is str:
UnboundLocalError: local variable 'level' referenced before assignment
```

pyspecdata/general_functions.py
jmfranck: `logger.INFO` and `logger.DEBUG` are.  But the `level` argument passed to this function is not (this is just a convenience function, and that's part of the point -- to be able to avoid importing logger yourself and also having to type out `logger.INFO`

examples/log_test.py
jmfranck: really, this works, but I'm trying to make sure it's worthwhile in terms of what you get out of it, so once everything is clear to you for a conversation, go ahead and resolve.  I will wait to merge until you resolve

pyspecdata/general_functions.py
guinneak198: gotcha

examples/log_test.py
guinneak198: this code is not working for me. will you take a look at my last merge and make sure everything is up to snuff? I don't understand why it's not working... error message is in previous comment

pyspecdata/log_test.py
jmfranck: My last commit actually worked.  I'm confused by what exactly you did.  It looks like you copied some but not all my code without actually pulling in the changes -- I will shoot you an image of what I see in gitk on slack

examples/log_test.py
jmfranck: the problem was that you mixed up `level`, which is what I called it and `stringLevel`, which is what you had called it

PR #11, opened 2020-05-28
generate a unittest for the new logging functionality

jmfranck:


jmfranck:
I'm replying to your comment above, but there's no reply button.

In general, you there are tests where you can check the value of arrays, etc (I think I posted the code from the guy in Songi's lab -- Yanxian -- which does that)

However, here, we are actually interested in testing *what the script prints out* since I'm trying to test the logger.  As it turns out, that's actually more complicated than the usual case, b/c you have the capture the output, etc.

PR #12, opened 2020-05-28
suggestion to avoid the "file being used by another process" issue

jmfranck:
THIS CODE HAS NOT BEEN TESTED


guinneak198 (commented):
So I am not sure what exactly you wanted me to review. I am going over general_functions.py and it looks like these are definitions for creating the documentation in sphinx? specifically process_kwargs, autostringconvert...
Is check_ascending_axis called in our FT function?
I see init_logging is defined here which I know to be very useful and important!
Kind of a computer science basics questions but why not make a class? Do we already have too many classes?

jmfranck:
@guinneak198 : this has nothing to do with sphinx -- click the files changed tab above to see what the pull request is doing. You'll see it's changing the `init_logging` function.  In particular, this should get around the problem where it tells you it's not able to open the log file (the reason you and @aabeaton were avoiding using logging).

I'm suggesting that you quickly:

-    Look through the changes to see what I did + that this makes sense to you.
-    Test it out -- specifically, try to run two scripts simultaneously that are both trying to log to a file.  (You should get multiple `pyspecdata.[number].log` files)
-    Debug, suggest changes, or ask questions as needed.

guinneak198:
Okay so I am not super familiar with logging however, I did run proc_CPMG and prc_Hahn_echoph with debug statements and it ran fine. I went into my pyspecdata.log file to check everything went smoothly and it looks okay to my amateur eye! I guess my question would be about the millions of font debug statements? I didn't put in anything about debugging the fonts. But there are screens on screens of the different fonts (i.e. ```DEBUG: findfont: score(<Font 'Arial' (ariblk.ttf) normal normal black normal>) = 10.525
--> font_manager.py(1271):matplotlib.font_manager     _findfont_cached 2020-09-10 10:56:12,360```) I have no idea if this is supposed to happen? My actual debug statement kind of gets drowned out. I also do not see the name of the file that I am debugging when I debug both... However, I now see this is for pyspecdata.log. When I go into pyspecdata.0.log it is up to date and does not drown out the info with the fonts

jmfranck:
what you mention are debugging statements that are apparently coming from matplotlib.  Apparently, it is [possible to suppress these](https://stackoverflow.com/questions/35325042/python-logging-disable-logging-from-imported-modules#:~:text=fileConfig()%20method%20accepts%20a,i.e.%2C%20in%20imported%20modules).&text=This%20disables%20all%20existing%20loggers,to%20load%20an%20external%20file).)

to, clarify, though, you run the two simultaneously, and you get two differently numbered `~/pyspecdata.#.log` files?

guinneak198:
yes specifically I see 
```$ ls *.log
pyspecdata.0.log  pyspecdata.1.log  pyspecdata.log```

jmfranck:
@guinneak198 : I went to merge this, and there were conflicts that might affect function.  I merged in master and resolved the conflicts -- can you please run your test again?

guinneak198:
Hmmm this doesn't work on master... I will run through the different commits and see which works. To clarify this is the error I get 
```Traceback (most recent call last):                                                                                        File "proc_Hahn_echoph.py", line 8, in <module>                                                                           logger = init_logging('debug')                                                                                        File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\general_functions.py", line 154, in init_logging              os.remove(log_filename)                                                                                             PermissionError: [WinError 5] Access is denied: 'C:\\Users\\Alex Guinness\\pyspecdata.log'    ```

guinneak198:
This commit still works 14542bbfa074a89b66f83be3a003c584fadc1a26

jmfranck:
OK, then the goal is to try to debug without rolling back.

jmfranck:
PS triple backticks to paste a code block -- use code blocks for tracebacks, not inline code

guinneak198:
I may not have pulled last time but when I git pulled today there were a lot of changes and when I tested again it produced the two log files numbered accordingly. So it seems this now does work without any edits needed!

PR #13, opened 2020-06-04
working on oom_values and such

guinneak198:
seems to work


pyspecdata/core.py
```diff
@@ -130,7 +130,14 @@ def apply_oom(average_oom,numbers,prev_label=''):
             raise ValueError(strm("you passed",average_oom,"which I can't find a prefix for"))
     else:
         oom_index = nonzero(eq)[0][0]
-    numbers[:] /= 10.**oom_values[oom_index]
+    print('type of numbers',numbers.dtype)
+    orig_dtype = numbers.dtype
+    oom = oom_values[oom_index]
+    print('type of oom', oom.dtype)
+    numbers = numbers.astype(float)
```
jmfranck, line 137: so, this actually won't work, b/c it creates a copy of numbers, and the point of the function seems to be to modify the `numbers` array in-place

jmfranck: (if you tried out this function by itself, you would see that it returns the prefix, but the numbers aren't scaled -- this will create a problem with the units)

pyspecdata/core.py
```diff
@@ -130,7 +145,15 @@ def apply_oom(average_oom,numbers,prev_label=''):
...
     else:
         oom_index = nonzero(eq)[0][0]
-    numbers[:] /= 10.**oom_values[oom_index]
+    if numbers.dtype in ['int32','int64']:
+        # this is not necessary unless we have an integer type
+        logger.warning("you are trying to determine the SI prefix of a"
+                "set of numbers that are described by integers.  This is"
+                "probably not a good idea!!")
+        new_values = numbers / 10.**oom_values[oom_index]
+        numbers[:] = new_values.astype(numbers.dtype)
```
jmfranck, line 154: notice I did the same thing you did, where we generate an array consisting of floats, but in this line, I slot them back into the same place in memory as before

PR #21, opened 2020-08-28
allow bruker parameters to be integers (using same rules as python)

jmfranck:


pyspecdata/load_files/bruker_nmr.py
```diff
@@ -262,7 +262,7 @@ def convert_to_num(val):
         elif '.' in val:
             return double(val)
         else:
-            return int(val)
+            return int(float(val))
```
jmfranck: can you give an example where this is required?  -- e.g. if I try `int(32)` in ipython, it generates the expected result.

guinneak198: this was done in regards to https://jmfrancklab.slack.com/archives/CV553TPRV/p1598553066001600

jmfranck: that was the reason that I created the pull request in the first place, yes.  But I'm asking -- why did you need to make the change that you made above?  Doesn't it work without this change?  If not, can you give an example where it does not work without this change?  Specifically, in the conditional, I'm testing to see whether or not it's a float, so I'm confused why this would be required here.

guinneak198: No it did not work I think we figured out that bruker loads some parameters in weird NaN ways? The error I discuss in that post is related to when it is set to `return int(val)`.  when I use this instead of `return int(float(val))` I still get an error 
`Traceback (most recent call last):
  File "proc_T1CPMG.py", line 23, in <module>
    expno=nodename,lookup=postproc_dict, fl=fl)
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\load_files\__init__.py", line 223, in find_file
    expno=expno)
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\load_files\__init__.py", line 390, in load_indiv_file
    data = bruker_nmr.series(file_reference, expno_as_str, dimname=dimname)
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\load_files\bruker_nmr.py", line 71, in series
    v = load_acqu(file_reference, *subpath)
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\load_files\bruker_nmr.py", line 254, in load_acqu
    return load_jcamp(file_reference,*subpath)
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\load_files\bruker_nmr.py", line 302, in load_jcamp
    data = list(map(convert_to_num,data))
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\load_files\bruker_nmr.py", line 265, in convert_to_num
    return int(val)
ValueError: invalid literal for int() with base 10: '2e-005'`
Albeit it is not the same error. But when I run it with the `return int(float(val))`  this error does not occur.

jmfranck: this is what I was asking for, thanks
(ps, use triple backticks to start and end a code block)

jmfranck: OK! @guinneak198  note that your patch doesn't actually fix this case, b/c:
```
In [2]: int(float('2e-5'))
Out[2]: 0
```
make sure my most recent commit (40d20d9) makes sense to you, and see if it works.

guinneak198: I see what you are saying, yes this makes sense and does still run well if not better!

PR #22, opened 2020-09-04
fixing extend

aabeaton:


pyspecdata/core.py
```diff
@@ -5458,6 +5458,14 @@ def extend(self,axis,extent, fill_with=0, tolerance=1e-5):
...
         start_index = 0
         stop_index = len(u)
+        stop_index = extent 
+        logger.debug("*** *** *** *** ***")
+        logger.debug(extent)
+        logger.debug(u[-1])
+        logger.debug(du)
+        logger.debug(extent*du-u[-1])
+        logger.debug(int((extent - u[-1]) // du))
+        logger.debug("*** *** *** *** ***")
```
jmfranck, line 5468: in general, try to use logger statements -- I know about the file writing issue, but I have a patch for that that I just want @guinneak198 to review

pyspecdata/core.py
```diff
@@ -5466,7 +5474,10 @@ def extend(self,axis,extent, fill_with=0, tolerance=1e-5):
             stop_index_addto = int((extent - u[-1]) // du)
             if ((extent - u[-1]) - du * stop_index_addto)/du > tolerance:# the first quantity here is negative
                 stop_index_addto += 1
+            if stop_index_addto > 100:
+                stop_index_addto = 1
```
jmfranck, line 5478: why is there a limit here?

pyspecdata/core.py
```diff
@@ -5466,7 +5474,10 @@ def extend(self,axis,extent, fill_with=0, tolerance=1e-5):
...
             stop_index += stop_index_addto
+            logger.debug(stop_index)
         else:
             raise RuntimeError("extent ({:g}) needs to be further than the bounds on '{:s}', which are {:g} and {:g}".format(extent,axis,u[0],u[-1]))
```
jmfranck, line 5482: a general comment -- can you add a **very** short file to the examples directory (under the main root of the repo) in order to give an example of the two use cases that you refer to here?  I would suggest, e.g., constructing an nddata of random numbers and extending it.  Show an example like you tried to start with and then an example like the "other case" you refer to (I imagine these are two differently shaped nddata).

examples/extend.py
```diff
@@ -0,0 +1,17 @@
...
+        'y':r_[0:0.1:100j]})
+with figlist_var() as fl:
+    fl.next('random data')
+    fl.image(d)
+    d.extend('x',1.5)
+    fl.next('extend along $x$')
+    fl.image(d)
+    d.extend('y',-0.05,fill_with=1)
+    fl.next('extend along $y$')
+    fl.image(d)
```
jmfranck, line 17: so, I added this example, whih I really thought should work, but it gives me some type of type error.  Can you edit either the example or the function?

pyspecdata/core.py
jmfranck: so I've attempted this with `examples/extend.py` above.

pyspecdata/core.py
jmfranck: I would reiterate this -- I don't understand why there's a limit here.

jmfranck (changes requested):
@aabeaton : I think we should be able to merge this in -- please check out comments in this review.

pyspecdata/core.py
aabeaton: This has been removed

pyspecdata/core.py
aabeaton: Yes, and this currently works as intended

examples/extend.py
aabeaton: This currently works

PR #24, opened 2020-09-04
debugging another issue with extend

aabeaton:


jmfranck:
this is already covered in #22

PR #29, opened 2021-02-12
add two examples from lmfit that we want to adapt

jmfranck:


guinneak198 (commented):
where are we on this? Last I checked I kind of had the multiple dataset running but weird x axis issues.

jmfranck:
This is super old, and I think superceded by the lmfitdata class

PR #32, opened 2021-05-14
try fix for ppm conversion

jmfranck:


examples/ppm_example.py
```diff
@@ -0,0 +1,43 @@
...
+TMS_shift = 12.5
+for A, nu, R in [
+    (0.1, TMS_shift, 0.04),
+    (1, 25, 0.08),
+    (1, 50, 1.2),
+    (1, 20, 0.45),
+    (1, 750, 1.2),
+    (1, 10, 0.08),
+]:
+    nu = SFO1 + nu # it's really unclear why this is done!
```
jmfranck, line 20: this doesn't make sense to me

examples/ppm_example.py
```diff
@@ -0,0 +1,43 @@
...
+# make peaks on a frequency axis
+x = nddata(r_[0:8:49152j], "t2")
+x.set_units("t2", "s")
+
+# generate time-domain signal
+signal = 0 * x  # create an array of zeros that's the right shape
+SFO1 = 400.130438262389
+TMS_shift = 12.5
+for A, nu, R in [
+    (0.1, TMS_shift, 0.04),
```
jmfranck, line 13: say that this represents my TMS peak -- how would I adjust my parameters to set this to 0 ppm?

examples/ppm_example.py
```diff
@@ -0,0 +1,43 @@
...
+    (1, 750, 1.2),
+    (1, 10, 0.08),
+]:
+    nu = SFO1 + nu # it's really unclear why this is done!
+    signal += A * exp(1j * 2 * pi * nu * x - x / R)
+signal.set_units("t2", "s")
+signal.ft("t2", shift=True)
+
+# Parameters needed by to_ppm to work
+signal.set_prop('acq',{'SFO1':400.130438262389,'SW_h':4807.69230769231,'O1':438.262389})
```
jmfranck, line 26: where do these numbers come from?

jmfranck: Does SW_h match 1/dt for the time axis above?

examples/ppm_example.py
jmfranck: if the above represent a table of offsets, why add in SFO1 here?

examples/ppm_example.py
```diff
@@ -0,0 +1,43 @@
...
+signal.set_prop('proc',{'OFFSET':12.50096})
```
jmfranck, line 27: the manual implies that this is `(SFO1/SF-1) * 1.0e6 + 0.5 * SW * SFO1/SF` for some acquisition modes, and without the second term for other acquisition modes -- what is SF, and is that true here? (I think SF is BF1, and so would depend on the dataset where these parameters are coming from)

jmfranck: If SF is BF1, then the first term is equivalent to `(SFO1-BF1)*1e6/BF1=O1/BF1` so that would be the offset from BF1 in ppm.  The second term I think is supposed to be half SW in units of ppm --> I think we actually see the difference in "offset" for the two modes if you look at the Hz axis for 2H vs. 1H -- the former has 0 on one side and for the other it's centered.

jmfranck: ahh -- see [here](http://www2.chem.uic.edu/nmr/downloads/processing-reference.pdf) pg 259 (search "basically makes BF1 = SF") which seems to imply that SF is related to BF1 using the parameter SR

examples/ppm_example.py
aabeaton: from a manual for Bruker 
image
-- I believe TMS is defined as the peak location where offset is 0, or where SFO1 = BF -- we cannot touch the BF, just the offset. It is still not extremely clear what happens. But to answer the question, that is you want this peak to represent TMS, wouldn't you just set TMS_shift = 0 ? For TMS, there would be no shift for it relative to its own peak, so a shift of 0 would just be the SFO1 which is consistent with what I am saying based off the Bruker manual

examples/ppm_example.py
aabeaton: I do not know why I thought I needed this, because it is fine without it. I will remove it.

examples/ppm_example.py
jmfranck: this is not correct --> TMS is rarely if ever on resonance with the carrier peak.  Turns out that the parameter I'm asking about is `SR` --> definitely check it out to cross-check.  More on this in next reply.

examples/ppm_example.py
jmfranck: aha!  This is related to the above question --> if you look up the definition of `SR`, it looks like `SF=SR+BF1` and it defines `SF` as the absolute frequency of the TMS.

see [p35 + p37 here](https://www.pascal-man.com/pulseprogram/avance3/topspin_2_1/proc_commands_parameters.pdf) and compare to our manual please.

The remaining question for our purposes is then which of these (`OFFSET`, `BF`, `SR`, `SF`) is actually stored, so that we can include the TMS referencing.

aabeaton: BF is an acquisition parameter and is stored.
OFFSET, SR, and SF are all stored in processing parameters.

pyspecdata/core.py
```diff
@@ -5242,15 +5242,15 @@ def to_ppm(self):
...
             self.ft('t2', shift=True) # this fourier transforms along t2, overwriting the data that was in self
         self.setaxis('t2', lambda x:
-                x/sfo1).set_units('t2','ppm')
-        max_ppm = (self.get_prop('acq')['SW_h']
-                / sfo1)
+                x-tms_hz).set_units('t2','ppm')
         self.setaxis('t2', lambda x:
-                (x-max_ppm+offset))
+                x/SR).set_units('t2','ppm')
         self.set_prop('x_inverted',True)
```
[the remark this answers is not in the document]
jmfranck: so -- follow up -- in the subsequent commit I actually did go ahead and test, and also fixed an SR/SF confusion that I had

PR #35, opened 2021-07-08
Create a pull request with only a bounded fit

jmfranck:


doc_src/conf.py
```diff
@@ -59,13 +59,13 @@
...
-            },
-        "doc_module": ("pyspecdata",),
+        #'image_srcset': ['2x'],
+        #'reference_url': {
+        #    'pyspecdata': None,
+        #    },
+        #"doc_module": ("pyspecdata",),
         ## directory where function/class granular galleries are stored
-        'backreferences_dir'  : 'gen_modules/backreferences',
+        #'backreferences_dir'  : 'gen_modules/backreferences',
```
jmfranck, line 68: don't do this

examples/DCCT_plotting.py
```diff
@@ -0,0 +1,29 @@
...
+This example creates a fake inversion recovery
+data set and plots the data with coherence
+transfer pathways in a reasonable way using
+DCCT plotting."""
+from pyspecdata import * 
+import numpy as np
+from pyspecdata.plot_funcs.DCCT import DCCT
+fl=figlist_var()
+t_axis = nddata(r_[0:2:256j],'t2')
+vd = nddata(r_[0,0.05,0.1,0.15,0.2,0.3,0.4,0.6,0.8,1.0,2.0,3.0,5.0],'vd')
```
jmfranck, line 13: unrelated

jmfranck: also, this has a quit statement, which breaks sphinx, which means that you couldn't have run sphinx to check!!

examples/fit_with_bounds.py
```diff
@@ -0,0 +1,143 @@
...
+logger.info(strm("p_true is:", p_true))
+x_vals = linspace(0, 250, 1500)
+empty_data = nddata(x_vals, "x").copy(data=False)
+# }}}
+# {{{making sympy expression
+A, shift, period, decay, x = sp.symbols("A shift period decay x")
+expr = A * sp.sin(shift + x / period) * sp.exp(-((x * decay) ** 2))
+# seems likely that Parameters is an ordered list, in which case, we don't need
+# parameter_names -- **however** we need to check the documentation to see that
+# this is true
```
jmfranck, line 94: @guinneak198 please do this!

jmfranck: (check the lmfit documentation to see if Parameters is an ordered list -- if it is, you don't need to return `parameter_names` below)

jmfranck: so, actually I went ahead and checked -- it is:

> In lmfit, this one-dimensional array is replaced by a Parameters object, which works as an ordered dictionary of Parameter objects with a few additional features and methods.

examples/fit_with_bounds.py
```diff
@@ -36,6 +36,7 @@ def gen_from_expr(expr, guesses={}):
     """
     # {{{ decide which symbols are parameters vs. variables
     all_symbols = expr.atoms(sp.Symbol)
+    # you need to pass "empty_data" to the function to enable the following line to work
```
jmfranck, line 39: this needs to be fixed

guinneak198: added definition of empty data inside function

examples/fit_with_bounds.py
```diff
@@ -18,78 +18,92 @@
...
+            pars.add(this_name, **kwargs)
+        for j in pars:
+            logger.info(strm("fit param ---", j))
+        logger.info(strm(pars))
+        self.fn = lambdify(
+            variable_symbols + parameter_symbols,
+            expr,
+            modules=[{"ImmutableMatrix": np.ndarray}, "numpy", "scipy"],
+        )
+        return pars
```
jmfranck, line 80: the *only* thing I change after wrapping it in a class  is this -- see how I'm storing the function as an attribute of the class rather than returning it -- explained more below.

examples/fit_with_bounds.py
```diff
@@ -18,78 +18,92 @@
...
+
+
+    def residual(self, pars, x, data=None):
+        "calculate the residual OR if data is None, return fake data"
+        shift = pars["shift"]
+        if abs(shift) > pi / 2:
+            shift = shift - sign(shift) * pi
+        model = self.fn(x, **pars.valuesdict())
```
jmfranck, line 88: before, this was just using the `myfunc` variable, which is actually defined in a global scope ([see here for definition/explanation](https://www.datacamp.com/community/tutorials/scope-of-variables-python)) in the code **underneath** which is *super confusing*.  So, I decided that there was no time like the present to start migrating to a class.

examples/fit_with_bounds.py
```diff
@@ -18,78 +18,92 @@
...
+thisfit = myfitclass()
+# {{{ a lot of what's below depends on knowing what the shape and dimension labels of my data are, so define that here
+x_vals = linspace(0, 250, 1500)
+empty_data = nddata(x_vals, "x").copy(data=False)
 # }}}
 # {{{making sympy expression
 A, shift, period, decay, x = sp.symbols("A shift period decay x")
 expr = A * sp.sin(shift + x / period) * sp.exp(-((x * decay) ** 2))
-fit_params, myfunc = gen_from_expr(
+fit_params = thisfit.gen_from_expr(
```
jmfranck: it was no longer necessary to return the lambda function, since it's now stored as an attribute of the class.

examples/fit_with_bounds.py
```diff
@@ -18,78 +18,92 @@
...
     expr,
-    {
+    empty_data,
```
jmfranck, line 105: `gen_from_expr` needs the info about the dimensions from the data -- again, it was pulling this out of the global scope, from the variable defined underneath, which was difficult to follow.

examples/fit_with_bounds.py
```diff
@@ -18,78 +18,92 @@
...
+        if abs(shift) > pi / 2:
+            shift = shift - sign(shift) * pi
+        model = self.fn(x, **pars.valuesdict())
+        if data is None:
+            return model
+        return model - data
+
+
+# }}}
+thisfit = myfitclass()
```
jmfranck, line 95: I need to instantiate the class to use it.

examples/fit_with_bounds.py
```diff
@@ -18,78 +18,92 @@
...
+# {{{ a lot of what's below depends on knowing what the shape and dimension labels of my data are, so define that here
+x_vals = linspace(0, 250, 1500)
+empty_data = nddata(x_vals, "x").copy(data=False)
 # }}}
 # {{{making sympy expression
 A, shift, period, decay, x = sp.symbols("A shift period decay x")
 expr = A * sp.sin(shift + x / period) * sp.exp(-((x * decay) ** 2))
```
jmfranck, line 102: so -- as an exercise, maybe try to change this: rather than defining as a global variable and passing to the function below, can you rather set an attribute of `thisfit` here, an then change `gen_from_expr` so that it doesn't require an expression argument?

jmfranck:
I did verify that sphinx builds and that there are no more global variables left inside the functions.

jmfranck:
So, in order to merge this, just deal w/ most of the comments above, I think all but one of which involve just giving thumbs up or adding needed questions.

examples/fit_with_bounds.py
guinneak198: make the expression an attribute of `thisfit`? but wouldn't that not be a good idea if we wanted to fit to a different expression it would be hardcoded in the class?

examples/fit_with_bounds.py
```diff
@@ -19,12 +19,8 @@
 np.random.seed(15816)
 # {{{ helper function(s)
 class myfitclass (object):
-    def this_expression(self):
-        A, shift, period, decay, x = sp.symbols("A shift period decay x")
-        expr = A * sp.sin(shift + x / period) * sp.exp(-((x * decay) ** 2))
-        return expr
-
-    def gen_from_expr(self, expr, data, guesses={}):
+    expression = (2+2)
```
jmfranck, line 22: you don't want to initialize the expression to something random.

There are different ways of approaching this, but I would go like this:
```
def __init__():
    self.expression = None
```
then below, in `gen_from_expr`, you would add a line at the beginning:
```
    if self.expression is None: raise ValueError("Ahhh! But you haven't set the expression yet!!")
```

guinneak198: okay this makes sense! will do!

PR #36, opened 2021-07-16
fix the convolution so that it's axis-aware

jmfranck:


aabeaton (commented):
I am cool with this

pyspecdata/fourier/convolve.py
```diff
@@ -25,11 +25,17 @@ def convolve(self,axisname,filterwidth,convfunc = (lambda x,y: exp(-(x**2)/(2.0*
...
+        print("detect self in frequency domain")
         self.ift(axisname)
         time_domain = False
     elif self.get_ft_prop(axisname,['start','freq']) is None:
+        # detect self in time domain, never FT'd
+        print("detect self in time domain, never FT'd")
         self.ft(axisname, shift=True)
     else:
+        # detect self in time domain, already FT'd
+        print("detect self in time domain, already FT'd")
```
jmfranck: log, don't print

pyspecdata/fourier/convolve.py
```diff
@@ -32,7 +32,7 @@ def convolve(self,axisname,filterwidth,convfunc = (lambda x,y: exp(-(x**2)/(2.0*
     else:
         self.ft(axisname)
     x = self.fromaxis(axisname)
-    myfilter = convfunc(x,filterwidth)
+    myfilter = convfunc(x,1/filterwidth)
     newdata = self*myfilter
```
jmfranck: this is explained in the commit message -- the filter width is given in the CURRENT domain

jmfranck: (That's why it's inverted)

pyspecdata/fourier/convolve.py
```diff
@@ -11,11 +13,19 @@ def convolve(self,axisname,filterwidth,convfunc = (lambda x,y: exp(-(x**2)/(2.0*
         apply the convolution along `axisname`
 
     filterwidth: double
-        width of the convolution function.
+        width of the convolution function (specify
+        in the conjugate domain, since this is
+        where the convolution function is
+        constructed)
+
```
jmfranck, line 20: this is actually not correct.

jmfranck: see my comment above about `1/fwhm`

pyspecdata/fourier/convolve.py
```diff
@@ -11,11 +13,19 @@ def convolve(self,axisname,filterwidth,convfunc = (lambda x,y: exp(-(x**2)/(2.0*
...
 
     convfunc: function
-        A function that takes two arguments -- the first are the axis coordinates and the second is `filterwidth`.
-        Default is a normalized Gaussian of width (:math:`\sigma`)
+        A function defined in the conjugate
+        Fourire domain of self that takes two
+        arguments -- the first are the axis
+        coordinates and the second is
+        `filterwidth`.  Default is a normalized
+        Gaussian of width (:math:`\sigma`)
```
jmfranck, line 28: I would roll this back

pyspecdata/fourier/convolve.py
```diff
@@ -38,7 +48,7 @@ def convolve(self,axisname,filterwidth,convfunc = (lambda x,y: exp(-(x**2)/(2.0*
         print("detect self in time domain, already FT'd")
         self.ft(axisname)
     x = self.fromaxis(axisname)
-    myfilter = convfunc(x,1/filterwidth)
+    myfilter = convfunc(x,filterwidth)
```
jmfranck: I would roll this back -- I actually changed it to `1/filterwidth` two commits back, as explained elsewhere.  If you don't do this, it breaks the old behavior of the convolve function.

pyspecdata/fourier/convolve.py
aabeaton: If I invert the filter width I get this, which does not show a difference in the convolution 210730_diagnostic1

aabeaton: whereas using just filterwidth, I get this, which shows a difference
210730_diagnostic2

aabeaton: wait I got mixed up - I think I understand this now... I will post back soon

aabeaton: ok I have got it now. disregard above.

pyspecdata/fourier/convolve.py
aabeaton: we are having a discussion on Slack, but for now I will keep this as `1/filterwidth`

aabeaton:
Ok I have addressed all parts to this. Should I merge with master at this point?

pyspecdata/fourier/convolve.py
jmfranck: I end up doing what I suggest on slack -- please check out and see that it works OK for you.

pyspecdata/fourier/convolve.py
jmfranck: yes, I have an update below

examples/indiv_funcs/convolve.py
```diff
@@ -11,14 +11,16 @@
...
 ax_time.set_title("time domain")
 ax_freq.set_title("frequency domain")
 fake_data.ft("t2", shift=True)
 assert fake_data.get_ft_prop("t2", "dt") == 1e-3
 plot(fake_data, ax=ax_freq, alpha=0.5, label="raw data")
-fake_data.convolve("t2", 0.07 / 2.355)
+# note that the convolution is done in the
+# frequency domain
+fake_data.convolve("t2", 10 / 2.355)
 plot(fake_data, ax=ax_freq, alpha=0.5, label="after convolve")
```
jmfranck: do run these demos -- my old convolution was too short.

jmfranck:
please check that this works for your purposes, and that the changes I'm making make sense to you -- if so, then I think we can merge.

pyspecdata/fourier/convolve.py
```diff
@@ -1,12 +1,12 @@
...
+from pylab import r_,fft,ifft,ifftshift,fftshift,exp,ones_like,sqrt,pi
 from pyspecdata import init_logging,strm
 logger = init_logging("debug")
 
-def convolve(self,axisname,filterwidth,convfunc = (lambda x,y: exp(-(x**2)/(2.0*(y**2))))):
-    r'''Perform a convolution on the data in its
-    current domain via multiplication in its
-    conjugate domain.
+def convolve(self,axisname,filterwidth,convfunc = (lambda x,y:
+    exp(-(x**2)/(2.0*(y**2)))/(y*sqrt(2*pi)))
```
aabeaton, line 7: I am not sure where this `(y*sqrt(2*pi))` comes from, otherwise this and the rest of the actual code of the function I agree with

jmfranck: that makes the gaussian normalized

jmfranck: (so that its integral is normalized)

PR #37, opened 2021-09-13
Lmfit class

guinneak198:
making lmfit (bounded) into class


examples/fit_with_bounds.py
```diff
@@ -18,21 +18,32 @@
 init_logging(level="debug")
 np.random.seed(15816)
 # {{{ helper function(s)
-class myfitclass (object):
-    def __init__(self):
+class myfitclass (nddata):
+    def __init__(self,*args,**kwargs):
+        # copied from fitdata
         self.expression = None
-    expression = (2+2)
```
guinneak198, line 24: will we still need a class variable saying expression = None? or that's taken care of inside of __init__?

examples/fit_with_bounds.py
```diff
@@ -18,21 +18,32 @@
...
-    def gen_from_expr(self,data, guesses={}):
+        if isinstance(args[0],nddata):
+            # move nddata attributes into the current instance
+            myattrs = normal_attrs(args[0])
+            for j in range(0,len(myattrs)):
+                self.__setattr__(myattrs[j],args[0].__getattribute__(myattrs[j]))
+        else:
+            nddata.__init__(self,*args,**kwargs)
+        return
+    @property
```
guinneak198, line 33: what are these @ symbols for?

examples/fit_with_bounds.py
```diff
@@ -18,21 +18,32 @@
...
+    def functional_form(self):
```
guinneak198, line 34: this function has the same name as the function below it? won't this cause issues?

examples/fit_with_bounds.py
```diff
@@ -43,15 +54,16 @@ def gen_from_expr(self,data, guesses={}):
         fn: function
```
guinneak198, line 54: the way we have it in master we only return pars?

examples/fit_with_bounds.py
```diff
@@ -61,58 +73,85 @@ def gen_from_expr(self,data, guesses={}):
...
-        for this_name in parameter_names:
-            kwargs = {}
-            if this_name in guesses.keys():
-                logger.debug(strm("applying bounds for", this_name))
-                kwargs.update(guesses[str(this_name)])
-            pars.add(this_name, **kwargs)
-        for j in pars:
-            logger.info(strm("fit param ---", j))
-        logger.info(strm(pars))
         self.fn = lambdify(
```
guinneak198, line 82: will this placement be okay? (before we declared Parameters prior to lambdify and now we declare Parameters after)

examples/fit_with_bounds.py
```diff
@@ -61,58 +73,85 @@ def gen_from_expr(self,data, guesses={}):
...
             variable_symbols + parameter_symbols,
             self.expression,
             modules=[{"ImmutableMatrix": np.ndarray}, "numpy", "scipy"],
         )
-        return pars
+        self.pars = Parameters()
+        for this_name in parameter_names:
+            self.pars.add(this_name)
+    def set_guess(self,*args,**kwargs):
```
guinneak198, line 90: in master we did this inside the main class method,  rather than making an entirely new method of the set_guesses. Do we want a new method for this? which do you think is the better option?

guinneak198 (commented):
AG questions/comments on lmfit class

examples/fit_with_bounds.py
jmfranck: How would you otherwise plan on setting the guess of the instance (this is not a rhetorical question — maybe you’re thinking of something I have not thought of)

examples/fit_with_bounds.py
jmfranck: The order of the methods inside the class definition has no meaning.

examples/fit_with_bounds.py
jmfranck: Are you commenting on the docstring here?  If so, it should just describe what the code does.

examples/fit_with_bounds.py
jmfranck: No — the decorator on the line before is different

examples/fit_with_bounds.py
jmfranck: They are decorators — good question.  If you look up a decorator tutorial, it will explain better than me.  Specifically here we are using the property decorator to create a property (which acts like an attribute but also does stuff behind the scenes when you set or get the value)

examples/fit_with_bounds.py
jmfranck: Inside init. We want an instance attribute.

examples/fit_with_bounds.py
guinneak198: guesses was a dictionary kwarg that one of the methods took in
```
def gen_from_expr(self,data, guesses={}):
    ...
    for this_name in parameter_names:
            kwargs = {}
            if this_name in guesses.keys():
                logger.debug(strm("applying bounds for", this_name))
                kwargs.update(guesses[str(this_name)])
            pars.add(this_name, **kwargs)

```

examples/fit_with_bounds.py
jmfranck: you are supposed to [edit the docstrings to describe the code](https://www.wrike.com/open.htm?id=748442764)

examples/fit_fake_data.py
```diff
@@ -25,12 +25,17 @@
...
+logger.info(strm("Functional Form", f.functional_form))
+logger.info(strm("Functional Form", f.functional_form))
+f.set_guess({M0:-500, Mi:500, R1:2})
 f.settoguess()
-fitdataguess = f.eval(100)
+guess = f.eval(100)
 f.fit()
+print("output:",f.output())
+print("latex:",f.latex())
+#}}}
```
jmfranck: there is no need for two separate sets of variable names here.

examples/fit_fake_data.py
```diff
@@ -54,16 +58,30 @@
 newfit.settoguess()
 newguess = newfit.eval(100)
 newfit.fit()
+# I don't understand -- minimize should be called as part of the fit
+# method, above -- why are you calling it explicitly here?
 out = minimize(
         thisfit.residual, newfit.pars, kws={"data":fake_data.data})
 newerfit = empty_data.copy(data=False)
 newerfit.data = newfit.residual(out.params)
```
jmfranck: see comment

examples/fit_fake_data.py
```diff
@@ -42,7 +47,6 @@
 # }}}
 #}}}
 #{{{lmfitdata method
-M0,Mi,R1,vd = sp.symbols("M0 Mi R1 tau")
```
jmfranck: again, no need to define the same thing twice

examples/fit_fake_data.py
```diff
@@ -54,16 +58,30 @@
...
+    text(0.5,0.5,f.latex(),
+            ha='center',va='center',
+            color=thisline[0].get_color(),
+            transform = ax.transAxes)
+    text(0.5,0.5,(3*'\n')+list_symbs,
+            ha='center',va='top',
+            size=10,
+            color=thisline[0].get_color(),
+            transform = ax.transAxes)
+    # }}}
```
jmfranck: this is an important feature -- don't delete it --> duplicate it

examples/fit_with_bounds.py
```diff
@@ -52,11 +52,13 @@
 guess = newfit.eval(100)
 # }}}
 # {{{ run the fit and generate nddata
+# again, now that this is a class, why is this not handled by the fit method?
 out = minimize(
     thisfit.residual, newfit.pars, kws={"data": mydata.data}
 )
 fit = empty_data.copy(data=False)
 fit.data = newfit.residual(out.params)
+# everything between this and the last comment should be part of the fit method, no?
```
jmfranck: see comments

examples/lmfitdata.py
```diff
@@ -1,3 +1,4 @@
+# just put this in the package
```
jmfranck, line 1: actually, I'll just do this

examples/fit_fake_data.py
```diff
@@ -21,7 +21,6 @@
 tau = nddata(r_[0:2:100j], 'tau')
 fake_data = 102*(1-2*exp(-tau*6.0))
 fake_data.add_noise(5.0)
-empty_data = nddata(r_[0:2:100j],'tau')
```
jmfranck: shouldn't be needed -- wasn't part of the code before, right?

examples/fit_fake_data.py
```diff
@@ -47,22 +46,20 @@
 # }}}
 #}}}
 #{{{lmfitdata method
-thisfit = lmfitdata(empty_data)
 newfit = lmfitdata(fake_data)
 newfit.functional_form = (Mi + (M0-Mi)*sp.exp(-(vd*R1)))
-thisfit.functional_form = newfit.functional_form
```
jmfranck: related to above

examples/fit_fake_data.py
```diff
@@ -47,22 +46,20 @@
...
 newfit.set_guess(
-        M0 = dict(value=-500, max=0, min=-501),
-        Mi=dict(value=500, max = 501, min=0), 
-        R1=dict(value=5, max = 6, min = 1))
+        M_0 = dict(value=-500, max=0, min=-501),
+        M_inf=dict(value=500, max = 501, min=0), 
+        R_1=dict(value=5, max = 6, min = 1))
```
jmfranck, line 54: it knows the names by their symbols, not by the variable names you've assigned those symbols to.

examples/fit_fake_data.py
guinneak198: I couldn't find the documentation in sympy for symbols to figure out what the real kwarg is but I know for lmfitdata it can't be in there but for fitdata it has to be in there...https://jmfrancklab.slack.com/archives/CV553TPRV/p1633962614001500

examples/fit_fake_data.py
guinneak198: I need this to initiate (instantiate?) the class with empty data before fitting the data into it. This is also how it's done in the fit with bounds example. Without it it does not fit appropriately.

examples/fit_fake_data.py
guinneak198: taken care of

examples/fit_fake_data.py
```diff
@@ -20,38 +20,44 @@
...
+            M_0 = dict(value=-500, max=0, min=-501),
+            M_inf=dict(value=500, max = 501, min=0), 
+            R_1=dict(value=5, max = 6, min = 1))
+    newfit.settoguess()
+    fl.plot(newfit.eval(100), label='lmfitdata guess')
+    # the following isn't equivalent to fitdata -- fit should be called
+    # without arguments, and it just runs minimize
+    # it should absolutely not take fake_data or the class as arguments, since it
+    # should already  know about it from the line `newfit=lmfitdata(fake_data)`
+    fit = newfit.fit(newfit,fake_data)
```
jmfranck, line 59: This is important -- I also comment on this inside the class

pyspecdata/lmfitdata.py
```diff
@@ -241,7 +241,7 @@ def thatresidual(self,p,x,y,sigma):
                     + explain_error(e))
         return retval
 
-    def fit(self, newfit, actualdata):
+    def fit(self):
```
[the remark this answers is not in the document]
jmfranck: "newfit" below I think can just be replaced by "self"

examples/fit_with_bounds.py
```diff
@@ -27,16 +27,12 @@
...
-for k, v in true_values.items():
-    p_true.add(k, value=v)
-logger.info(strm("p_true is:", p_true))
-# }}}
-# {{{ nddata to generate the fake data
-mydata = empty_data.copy(data=False)
-mydata.data = thisfit.residual(p_true)
+thisfit.set_guess(true_values)
+thisfit.settoguess()
+mydata = thisfit.output()
```
jmfranck: this should be possible, and shouldn't be changed.

pyspecdata/lmfitdata.py
```diff
@@ -160,18 +160,26 @@ def set_guess(self, *args, **kwargs):
...
+                elif np.isscalar(this_name):
+                    self.pars[this_name].value = guesses[this_name]
+                    self.guess_dict[this_name] = {"value":guesses[this_name]}
+                else:
+                    raise ValueError("what are the keys to your guesses???")
         for j in self.pars:
             logging.info(strm("fit param ---", j))
         logging.info(strm(self.pars))
-        self.guess_dict = guesses
         return
```
jmfranck: this was not working with the code above, so I'm starting to fix here.

pyspecdata/lmfitdata.py
```diff
@@ -279,22 +278,17 @@ def fit(self):
...
-            )
-            sigma = np.ones(np.shape(y))
-        min_args = (self.residual, self.pars)
-        min_kwargs={'args':(x,y,sigma)}
-        out = minimize(*min_args,**min_kwargs, kws={"data": self.data})
+        out = minimize(
+            self.residual,
+            self.pars,
+            args=(x, y, sigma),
+        )
```
jmfranck: just cleaning up

examples/fit_with_bounds.py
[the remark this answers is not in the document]
guinneak198: oh shoot yes I changed fit_fake_data for the updates but forgot to apply the changes to the fit_w_bounds!

pyspecdata/lmfitdata.py
```diff
@@ -307,20 +301,33 @@ def run_lambda(self, pars):
             *(self.getaxis(j) for j in self.variable_names), **pars.valuesdict()
         )
 
-    def residual(self, pars, x, y, sigma, data=None):
+    def residual(self, pars, x, y, sigma=None):
```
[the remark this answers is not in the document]
guinneak198: why?

examples/fit_fake_data.py
```diff
@@ -14,7 +14,15 @@
...
-
+def list_symbs(f):
+    # {{{ this is just to show all the parameters
+    list_symbs = []
+    for j,k in f.output().items():
+        s_repr = sp.latex(sp.Symbol(j))
+        list_symbs.append(f'${s_repr} = {k:0.5g}$')
+    list_symbs = '\n'.join(list_symbs)
+    # }}}
+    return list_symbs
```
jmfranck, line 25: I'm using this to add the annotations that you had deleted -- it seems to work

examples/fit_fake_data.py
```diff
@@ -66,19 +78,29 @@
...
         ha="center",
         va="center",
         color=thatline[0].get_color(),
         transform=ax.transAxes,
     )
+    text(0.6,0.5,(3*'\n')+list_symbs(newfit),
+            ha='center',va='top',
+            size=10,
+            color=thisline[0].get_color(),
+            transform = ax.transAxes)
```
jmfranck: So I decided on a whim to add these back in, and I found that they were actually broken, not eliminated by choice!

examples/fit_fake_data.py
```diff
@@ -3,57 +3,104 @@
...
-            color=thisline[0].get_color(),
-            transform = ax.transAxes)
+            color=thiscolor,
+            transform=ax.transAxes,
+        )
+
+    # {{{ use fitdata
+    fitdata_instance = fitdata(fake_data)
+    fitdata_instance.functional_form = functional_form
+    fitdata_instance.set_guess({M0: -500, Mi: 500, R1: 2})
```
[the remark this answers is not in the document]
jmfranck: b/c it doesn't matter

PR #44, opened 2023-04-13
Lmfit complex

guinneak198:
fit when the residual contains complex values


pyspecdata/figlist.py
```diff
@@ -64,7 +64,7 @@
 from .plot_funcs import image as this_image
 from .general_functions import *
 from . import plot_funcs as this_plotting
-from .core import nddata, autolegend, gridandtick
+from .core import nddata, autolegend, gridandtick, r_
```
guinneak198, line 67: just a minor import I know it's unrelated but I was getting an error when plotting that this fixed

jmfranck: from numpy, not .core

examples/fit_complex.py
```diff
@@ -0,0 +1,64 @@
+"""
+Fit Using Bounds
+================
+
+Adapt one of the examples from lmfit to use sympy and pyspecdata.
+Eventually, we want to use parts of this in the base library, so we don't have
+to retype them every time.
```
jmfranck, line 7: new docstring

pyspecdata/lmfitdata.py
```diff
@@ -100,40 +100,16 @@ def functional_form(self, this_expr):
...
         )
 
-        def fn(p, x):
-            p = self.add_inactive_p(p)
-            assert len(p) == len(
-                self.parameter_names
-            ), "length of parameter passed to fitfunc doesnt match number of symbolic parameters"
-            return self.fitfunc_multiarg(*tuple(list(p) + [x]))
-
-        self.fitfunc = fn
```
jmfranck, line 122: this add_inactive stuff is obsolete because lmfit has a mechanism for fixing parameters (which was the purpose of the "inactive parameter"

also fitfund *should not be used, since it is not how the residual calculates the fit function*

jmfranck: note that this change will **break** eval -- I know this -- do not fix eval by rolling back this change.

pyspecdata/figlist.py
guinneak198: okay added question on slack!

pyspecdata/lmfitdata.py
guinneak198: self.fitfunc is used later on line 256?

examples/fit_complex.py
```diff
@@ -20,7 +20,8 @@
 # create an empty dataset that we will drop the fake data into
 thisfit = lmfitdata(nddata(r_[-1:1:1001j], "t"))
 thisfit.functional_form = A * sp.exp(-1j*2*pi*nu*t) * sp.exp(-R*sp.pi*abs(t))
-only_real = True # if you set this to True, it works
+only_real = False # if you set this to True, it previously worked -- this
+#                   example demonstrates that this also works when set to False
```
jmfranck: just see that my comment here makes sense

pyspecdata/lmfitdata.py
```diff
@@ -283,9 +283,16 @@ def fit(self):
...
+        """actually run the lambda function that calculates the model data.
+        Note that the name of the variable along which the model data is calculated
+        (as opposed to "parameter" is set by variable_names parameter).
+
+        .. note::
+            we had separates this in case we want
+            our function to involve something else, as well (e.g. taking a Fourier
+            transform).  Unknown if there are still two steps in this way.
+
+        """
```
jmfranck, line 295: again, see that this makes sense

pyspecdata/lmfitdata.py
```diff
@@ -266,8 +242,13 @@ def eval(self, taxis=None, set_what=None, set_to=None):
...
         # }}}
-        newdata
-        newdata.data[:] = self.fitfunc(p, taxis).flatten()
+        param_dict = {}
+        for j in range(len(list(p))):
+            case = {list(self.parameter_names)[j]:list(p)[j]}
+            param_dict.update(case)
+        self.set_guess(param_dict)
+        self.fitfunc = self.run_lambda(self.pars)
+        newdata.data[:] = self.fitfunc.flatten()
```
jmfranck, line 251: the difference in your last line here is causing a problem you noted on slack.
You should read the eval function, and see how it uses and then modifies the variable `taxis`.
You'll see that `fitfunc` and therefore `run_lambda` needs to take taxis as an argument (if `run_lambda` is used elsewhere, it might make sense to make this an optional argument).

examples/fit_fake_data.py
```diff
@@ -30,7 +30,7 @@ def list_symbs(f):
 # }}}
 fl = figlist_var()
 # {{{creating a fake data recovery curve
-tau = nddata(r_[0:2:100j], "tau")
+tau = nddata(r_[0:2:256j], "tau")
```
jmfranck, line 33: what's the point of this?

pyspecdata/lmfitdata.py
```diff
@@ -266,8 +248,13 @@ def eval(self, taxis=None, set_what=None, set_to=None):
             newdata.set_error(self.fit_axis,
                     self.get_error(self.fit_axis))
         # }}}
+        #param_dict = {}
+        #for j in range(len(list(p))):
+        #    case = {list(self.parameter_names)[j]:list(p)[j]}
+        #    param_dict.update(case)
+        #self.set_guess(param_dict)
```
jmfranck, line 248: be sure to get rid of this

pyspecdata/lmfitdata.py
```diff
@@ -100,12 +99,6 @@ def functional_form(self, this_expr):
         )
         self.fit_axis = list(self.fit_axis)[0]
         # }}}
-        args = self.parameter_symbols + [str(*this_axis)]
-        self.fitfunc_multiarg = sp.lambdify(
-            args,
-            self.expression,
-            modules=[{"ImmutableMatrix": np.ndarray}, "numpy", "scipy"],
-        )
```
jmfranck, line 108: I deleted this stuff because it's better to use `_v2`, then you brought it back.  Please leave it deleted.

It's not got to have more than one thing that's trying to do the same thing.

The new format uses the `variable_symbols` attribute, while this does not.

Please use `fitfunc_multiarg_v2` in your new additions, which should be possible.

jmfranck: the ordering in `_v2` is different, and it's better to be consistent.  Also, we want to leave open the possibility that `variable_symbols` has more than one variable in it.  At the simplest level, you could just check that `variable_symbols` only had one element, and then feed `taxis` to the right argument of `multi...v2` -- for now it's OK to throw an error in the event that there is more than one variable.  (Down the line, we could just pass a dictionary to `taxis` if there is more than one.

jmfranck (changes requested):

examples/fit_fake_data.py
guinneak198: testing when I feed eval a number that is different from that of the data

pyspecdata/lmfitdata.py
```diff
@@ -266,8 +241,12 @@ def eval(self, taxis=None, set_what=None, set_to=None):
             newdata.set_error(self.fit_axis,
                     self.get_error(self.fit_axis))
         # }}}
-        newdata
-        newdata.data[:] = self.fitfunc(p, taxis).flatten()
+        param_dict = {}
+        for j in range(len(list(p))):
+            case = {list(self.parameter_names)[j]:list(p)[j]}
+            param_dict.update(case)
+        self.set_guess(param_dict)
```
jmfranck, line 248: I don't understand this -- why are you setting the guess here?  Couldn't this just be eliminated?

pyspecdata/lmfitdata.py
```diff
@@ -303,18 +280,31 @@ def fit(self):
...
+        if len(self.variable_names) > 1:
+            print("We are only allowing for one variable for now")
+        elif thistaxis is None:
+            return self.fitfunc_multiarg_v2(
+                    *(self.getaxis(j) for j in self.variable_names), **pars.valuesdict())
+        else:    
+            logging.debug(strm(self.getaxis(j) for j in self.variable_names))
+            return self.fitfunc_multiarg_v2(
+                    thistaxis, **pars.valuesdict()
+            )
```
jmfranck, line 303: yes, this is good!

jmfranck: let's modify, though, since we are thinking about this.
`thistaxis is None` to `len(kwargs) == 0` then in else, 
first argument is not `thistaxis` but `*(kwargs[j] if j in kwargs.keys() else self.getaxis(j) for j in self.variable_names)`

happy to talk about that to make sure it makes sense.

jmfranck: and `thistaxis=None` gets replaced by `**kwargs`

pyspecdata/lmfitdata.py
```diff
@@ -303,18 +280,31 @@ def fit(self):
...
 
     def residual(self, pars, x, y, sigma=None):
         "calculate the residual OR if data is None, return fake data"
-        fit = self.run_lambda(pars)
+        fit = self.run_lambda(pars,thistaxis = None)
```
jmfranck, line 307: you should be able to just leave this line alone (even before, this was your default, so there should be no need to change) --> this will not work with the change I suggest above.

pyspecdata/lmfitdata.py
```diff
@@ -266,8 +241,12 @@ def eval(self, taxis=None, set_what=None, set_to=None):
...
                     self.get_error(self.fit_axis))
         # }}}
-        newdata
-        newdata.data[:] = self.fitfunc(p, taxis).flatten()
+        param_dict = {}
+        for j in range(len(list(p))):
+            case = {list(self.parameter_names)[j]:list(p)[j]}
+            param_dict.update(case)
+        self.set_guess(param_dict)
+        newdata.data[:] = self.run_lambda(self.pars,thistaxis = taxis).flatten()
```
jmfranck, line 249: OK, so with what I suggest below, you would replace `thistaxis = taxis` in the call to `run_lambda` with `**{self.fitaxis:taxis}`

these changes are incomplete, but that way at least what we are proposing is compatible with more than one variable (most of what is/was there is compatible with only one variable, but think about fitting a 2D function like R1(τ,p))

pyspecdata/lmfitdata.py
jmfranck: sometimes, when we `eval` we do this after manually setting the parameters to the guess, but sometimes, we do this after fitting.  I'm suggesting that only the line added after this comment should be needed.

jmfranck (changes requested):
I want to be clear that what you had here was *good* but I'm pursuing a bit of a forward-looking agenda here.

pyspecdata/lmfitdata.py
jmfranck: (remember that we have a separate `settoguess` function for exactly this purpose)

jmfranck: so, to be clear, I'm suggesting removing this for loop and the subsequent `set_guess`

pyspecdata/lmfitdata.py
```diff
@@ -180,27 +180,20 @@ def _taxis(self, taxis):
...
-            the specific value(int) you are assigning the symbol you included
 
         Returns
         -------
         self: nddata
             the fit function evaluated along the axis coordinates that were passed
         """
-        if isinstance(set_what, dict):
-            set_to = list(set_what.values())
-            set_what = list(set_what.keys())
```
jmfranck, line 228: here, I'm just removing obsolete stuff (lmfit allows us to fix parameters by setting the `vary` property of a parameter to `False`)

pyspecdata/lmfitdata.py
```diff
@@ -209,17 +202,6 @@ def eval(self, taxis=None, set_what=None, set_to=None):
...
-        if set_what is not None:
-            if self.set_indices is not None:
-                raise ValueError(
-                    "You're trying to set indices in an eval"
-                    " function for a function that was fit constrained; this"
-                    " is not currently supported"
-                )
-            set_indices, set_to, active_mask = self.gen_indices(set_what, set_to)
-            p[set_indices] = set_to
-        # }}}
```
jmfranck, line 247: same thing here

pyspecdata/lmfitdata.py
guinneak198: sounds good

pyspecdata/lmfitdata.py
guinneak198: cool

pyspecdata/lmfitdata.py
guinneak198: https://jmfrancklab.slack.com/archives/CV553TPRV/p1682684466462009?thread_ts=1682650761.090889&cid=CV553TPRV

pyspecdata/lmfitdata.py
guinneak198: https://jmfrancklab.slack.com/archives/CV553TPRV/p1682685969709779?thread_ts=1682650761.090889&cid=CV553TPRV

pyspecdata/lmfitdata.py
guinneak198: I think you're saying "no need to change" and when I leave this both examples work

jmfranck: no, with suggestions below, your change will not work.  Leave the way it was before the PR (red line)

pyspecdata/lmfitdata.py
jmfranck: core of this issue is ongoing, addressed with new commit

pyspecdata/lmfitdata.py
```diff
@@ -223,9 +225,11 @@ def eval(self, taxis=None):
...
                     self.get_error(self.fit_axis))
         # }}}
-        for j,this_name in enumerate(self.pars.keys()):
-            self.pars[this_name].value = p[j]
-        newdata.data[:] = self.run_lambda(self.pars,**{self.fit_axis:taxis}).flatten()
+        for j,this_name in enumerate(self.guess_coeff.keys()):
+            # as discussed, guess_coeff are *not* the fit_coeff.  
+            # (I changed the name from "pars" to "guess_coeff" to clarify that, since "pars" was arguably a bad name!).
+            self.guess_coeff[this_name].value = p[j]
+        newdata.data[:] = self.run_lambda(self.guess_coeff,**{self.fit_axis:taxis}).flatten()
```
jmfranck, line 232: this is the core issue that needs to be addressed that I'm talking about in the rest of the review.  You are effectively setting the attribute that stores the guess to the fit result, and then running the function using the guess.  Why not just run using the fit result (attribute `fit_coeff`)?

jmfranck (changes requested):

pyspecdata/lmfitdata.py
jmfranck: since this was a tangential issue, I just did it in my new commit.

pyspecdata/lmfitdata.py
```diff
@@ -105,9 +105,9 @@ def functional_form(self, this_expr):
             modules=[{"ImmutableMatrix": np.ndarray}, "numpy", "scipy"],
         )
 
-        self.pars = Parameters()
+        self.guess_coeff = Parameters()
         for this_name in self.parameter_names:
-            self.pars.add(this_name)
+            self.guess_coeff.add(this_name)
```
jmfranck: throughout, I changed the name of `self.pars` to `self.guess_coeff` since that's a better name for what it is, and because I think this was causing you confusion.

PR #48, opened 2024-02-05
changes to fix lmfit version

wfkincaid:
Change to setup.py to require more updated version of lmfit, if lmfit is already installed


jmfranck:
I need to check that setup.py still works ok

PR #50, opened 2024-07-23
setting units based on parameters from hdf5 file!

guinneak198:
newer datasets show units are set in hdf5 but this wasn't being reflected properly upon loading


pyspecdata/core.py
```diff
@@ -7536,6 +7537,10 @@ def _init_datanode(self, datanode, **kwargs):
                         + explain_error(e)
                     )
                 recordarrayofaxis = datadict["axes"][axisname]["data"]
+                if "axis_coords_units" in datadict["axes"][axisname].keys():
+                    myaxis_units[axisnumber] = datadict["axes"][axisname].__getitem__("axis_coords_units")
```
jmfranck, line 7541: Just use square brackets

pyspecdata/core.py
```diff
@@ -7552,6 +7557,7 @@ def _init_datanode(self, datanode, **kwargs):
                     )
                 datadict["axes"].pop(axisname)
             kwargs.update({"axis_coords": myaxiscoords})
+            kwargs.update({"axis_coords_units": myaxis_units})
```
jmfranck, line 7560: To make sure that this is being done correctly, I am pretty sure that there is an example that writes and reads an hdf file.. this should involve modifying that example as well

pyspecdata/core.py
jmfranck, line 7541: This is not going to work in the general case. Unless you initialize this to a list of the correct length... I would start by initializing to a list of Nones (multiply a list of length 1 with element None by whatever length you want for the list)

pyspecdata/core.py
```diff
@@ -7536,6 +7537,10 @@ def _init_datanode(self, datanode, **kwargs):
                         + explain_error(e)
                     )
                 recordarrayofaxis = datadict["axes"][axisname]["data"]
+                if "axis_coords_units" in datadict["axes"][axisname].keys():
+                    myaxis_units[axisnumber] = datadict["axes"][axisname].__getitem__("axis_coords_units")
+                else:
+                    myaxis_units[axisnumber] = None
```
jmfranck, line 7543: If you do what I'm saying above, you don't need this

jmfranck (changes requested):

pyspecdata/core.py
```diff
@@ -7508,6 +7508,7 @@ def _init_datanode(self, datanode, **kwargs):
         if "axes" in list(datadict.keys()):
             myaxiscoords = [None] * len(mydimlabels)
             myaxiscoordserror = [None] * len(mydimlabels)
+            myaxis_units = [None] * len(mydimlabels)
```
guinneak198, line 7511: I think this handles what you are talking about [here](https://github.com/jmfrancklab/pyspecdata/pull/50/files#r1688471252)

pyspecdata/core.py
guinneak198: tested and passed

pyspecdata/core.py
guinneak198: it is initialized above

pyspecdata/core.py
guinneak198: yes I nixed the else

pyspecdata/core.py
jmfranck: Btw sorry I didn't see this comment:

Which file are we talking about here?  I'm specifically asking that you should modify the existing example so that it checks that it can see the units that were saved to file.

jmfranck: @guinneak198 can you please update the example, so that we can merge this PR??

examples/test_units.py
```diff
@@ -0,0 +1,13 @@
+""" Test units are transferred from HDF5 file
+=============================================
+Test that the units saved in the HDF5 file are transferred upon opening
```
jmfranck, line 3: please modify ``examples/indiv_funcs/hdf_test.py`` instead

PR #51, opened 2024-07-24
fix real/imag color binding

jmfranck:


wfkincaid:
This looks good and flake is already happy

pyspecdata/plot_funcs/pcolormesh.py
```diff
@@ -54,7 +53,7 @@ def pcolormesh(
             fig = plt.gcf()
             ax1 = fig.add_subplot(1, 1, 1)
         ax_list = [(ax1, lambda x: x.real, "real")]
-    X, Y = np.meshgrid(
+    X, Y = plt.meshgrid(
```
jmfranck, line 56: I might have said incorrectly before, but this ultimately comes from numpy -- so not changing this is better.

pyspecdata/core.py
```diff
@@ -578,9 +578,10 @@ def lookup_rec(A, B, indexpair):
             )
         )
     whichisindex = joined[0][1].dtype.names.index(indexpair[1])
-    allbutindex = (
-        lambda x: list(x)[0:whichisindex] + list(x)[whichisindex + 1 :]
-    )
+
+    def allbutindex(x):
+        return list(x)[0:whichisindex] + list(x)[whichisindex + 1 :]
+
```
jmfranck: it seemed that there was nothing in core.py that we would want, so I just reverted to the version that's currently in core (after flake8 compliance)

jmfranck (commented):
again, make sure this works for you -- I also gave figlist the same treatment as core

PR #52, opened 2024-07-24
Fix sqrt

wfkincaid:


pyspecdata/core.py
```diff
@@ -308,8 +308,24 @@ def mybasicfunction(first_figure=None):
...
+        return sympy_sqrt(arg)
+    else:
+        return np_sqrt(arg)
+||||||| d1d83c22
+    'tests if something is sympy (based on the module name)'
+    return isinstance(x,sympy_expr)
+=======
     "tests if something is sympy (based on the module name)"
     return isinstance(x, sympy_expr)
+>>>>>>> master
```
jmfranck, line 328: this conflict needs resolution -- you just left the conflict markers in.

jmfranck: the middle block is the starting point, and the thing on top vs. thing on bottom are the two conflicting changes to the middle block.  Typically, you want to do both.

e.g. here, the top block added the psd_sqrt function, while the bottom block black-formatted the strings.  So, you probably want something like the top block, but where the strings are black formatted.  Be sure to remove the conflict markers!

setup.cfg.windows
```diff
@@ -1,4 +0,0 @@
-[build]
-compiler=mingw32
-[build_ext]
-compiler=mingw32
```
jmfranck, line 4: do `git checkout master setup.cfg.windows` to restore this file.

pyspecdata/core.py
jmfranck: yeah, actually just give me a minute

jmfranck:
this is based on a bad merge, so I'm closing it and opening a new one

PR #56, opened 2024-09-02
Concat

guinneak198:
edit concatenation function so that if the dimension already exists we just concatenate instead of making a new dimension


pyspecdata/core.py
```diff
@@ -38,7 +38,12 @@
 import sympy as sp
 from numpy import r_, c_, nan, inf, pi
 from mpl_toolkits.mplot3d import axes3d
-import textwrap, atexit, scipy, warnings, inspect, re
+import textwrap
+import warnings
+import scipy
+import atexit
+import inspect
+import re
```
guinneak198, line 46: to pass flake

pyspecdata/core.py
```diff
@@ -1616,109 +1622,131 @@ def concat(datalist, dimname, chop=False):
...
-        # {{{ make list for the shape to check, which contains the dimensions we are NOT concatting along
-        if dimname in shapes[j].dimlabels:
-            newdimsize += shapes[j][dimname]
-            shapetocheck = list(shapes[j].shape)
-            shapetocheck.pop(shapes[j].axn(dimname))
-        else:
+    if dimname in datalist[0].dimlabels:
+        print(
+            "The dimname you specified exists already in one of the datasets so I am assuming it also exists in the other"
+        )
```
guinneak198, line 1628: unsure if needed but I figure useful to tell the user what's going on

pyspecdata/core.py
```diff
@@ -1616,109 +1622,134 @@ def concat(datalist, dimname, chop=False):
...
+        # {{{ make list of the data axes and dimname axes for datasets
+        ndarrays = []  # will contain numpy array of data for each dataset
+        conc_axis = [] # will contain numpy array of the dimname axis for each dataset
```
guinneak198, line 1631: these are the 2 things that need to be concatenated

pyspecdata/core.py
```diff
@@ -1616,109 +1622,134 @@ def concat(datalist, dimname, chop=False):
...
+        for j in range(len(datalist)):
+            ndarrays.append(datalist[j].data)
+            conc_axis.append(datalist[j][dimname])
```
guinneak198, line 1634: since I'm appending to both lists I just made a for loop

pyspecdata/core.py
```diff
@@ -1616,109 +1622,134 @@ def concat(datalist, dimname, chop=False):
...
+        #     dimname so there is a slight difference
+        extra_dims = False
+        if len(datalist[-1].axis_coords) > 1:
+            extra_dims = True
+            extra_dimlabels = list(datalist[-1].dimlabels)
+            extra_axis_coords = list(datalist[-1].axis_coords)
+            dimname_idx = extra_dimlabels.index(dimname)
+            extra_dimlabels.pop(dimname_idx)
+            extra_axis_coords.pop(dimname_idx)
+        # }}}
```
guinneak198, line 1654: inspired by lines 1731 below but I add up here since we don't want to include the newdimsize stuff that is done down there

pyspecdata/core.py
```diff
@@ -1616,109 +1622,134 @@ def concat(datalist, dimname, chop=False):
...
+        newdatalist = nddata(newdata, datalist[0].dimlabels)
+        newdatalist.setaxis(dimname, new_dimname).set_units(
+            dimname, datalist[0].get_units(dimname)
+        )
```
guinneak198, line 1658: makes sure units are set properly before returning

pyspecdata/core.py
```diff
@@ -1616,109 +1622,134 @@ def concat(datalist, dimname, chop=False):
...
+                                shapetocheckagainst,
+                                "where all the shapes of the things you're trying to concat are:",
+                                shapes,
+                            )
                         )
-                    )
-    newdatalist = ndshape(datalist[-1])
-    if dimname in newdatalist.dimlabels:
-        newdatalist[dimname] = newdimsize
-    else:
```
guinneak198, line 1671: handle the scenario of existing dimname above so remove here

pyspecdata/core.py
```diff
@@ -1616,109 +1622,134 @@ def concat(datalist, dimname, chop=False):
...
-    try:
-        newdatalist = newdatalist.alloc()
-    except Exception:
-        raise ValueError(
-            strm(
-                "trying to alloc the newdatalist",
-                newdatalist,
-                "created a problem",
-            )
-        )
```
guinneak198, line 1683: this is a weird diff but you can see it's just shifted below (indented)

pyspecdata/core.py
jmfranck: please check your myflake -- it should be
```
#!/usr/bin/bash
flake8 --inline-quotes='"' --max-line-length=79 --ignore=E203,E225,E226,E231,E261,E265,E293,E302,E305,E306,E401,E501,E731,E741,Q002,W291,W293,W503 $*
```

pyspecdata/core.py
jmfranck: replaced with assertion

examples/concatenate.py
```diff
@@ -13,14 +13,18 @@
...
     "t"
 )  # sort is already applied in function but included here for clarity
+print("concatenate after sort:")
 print(combined)
 # }}}
 # {{{ when we make a new dimension - demos the existing functionality -- see above about kwarg
 data1 = psd.nddata(r_[0:4:4j], "t")
 data2 = psd.nddata(r_[0:4:4j], ["t"]) * 1.3
 combined = psd.concat([data1, data2], dimname="indirect")
+print("concatenate along new dimension:")
```
jmfranck, line 28: I didn't want to create a huge diff, but please move this file to `examples_undoc`

pyspecdata/core.py
```diff
@@ -1623,43 +1618,32 @@ def concat(datalist, dimname, chop=False):
...
     other_info_out = datalist[0].other_info
     if dimname in datalist[0].dimlabels:
-        print(
-            "The dimname you specified exists already in one of the datasets so I am assuming it also exists in the other"
+        assert all(
+            [dimname in datalist[j].dimlabels for j in range(len(datalist))]
+        ), (
+            "the dimension '"
+            + dimname
+            + "' exists in the first datset, but not the second!"
```
jmfranck, line 1626: replacing with an assertion is better

pyspecdata/core.py
```diff
@@ -1623,43 +1618,32 @@ def concat(datalist, dimname, chop=False):
...
         # {{{ make list of the data axes and dimname axes for datasets
         ndarrays = []  # will contain numpy array of data for each dataset
-        conc_axis = [] # will contain numpy array of the dimname axis for each dataset
+        conc_axis = (
+            []
+        )  # will contain numpy array of the dimname axis for each dataset
         for j in range(len(datalist)):
             ndarrays.append(datalist[j].data)
             conc_axis.append(datalist[j][dimname])
+        # PR: much more compact to just do this with a list comprehension in place, below
```
jmfranck, line 1636: please see comment

pyspecdata/core.py
```diff
@@ -1623,43 +1618,32 @@ def concat(datalist, dimname, chop=False):
...
         # }}}
         # {{{ concatenate the data ndarrays and dimname axis ndarrays
-        newdata = np.concatenate(
+        concated_data = np.concatenate(
```
jmfranck, line 1639: I changed the names here -- not just for yours but for the stuff below, which I hadn't read in a while

pyspecdata/core.py
```diff
@@ -1623,43 +1618,32 @@ def concat(datalist, dimname, chop=False):
...
-        newdatalist.setaxis(dimname, new_dimname).set_units(
-            dimname, datalist[0].get_units(dimname)
-        )
-        newdatalist.sort(dimname)
-        if extra_dims:
-            # make sure we still have the indirect axes we started with
-            newdatalist.labels(extra_dimlabels[0], extra_axis_coords[0])
+        retval = datalist[-1].copy(data=False)
+        retval.axis_coords[datalist[-1].axn(dimname)] = concated_ax_coords
+        retval.data = concated_data
```
jmfranck, line 1646: note this is much simpler than above.  I think the stuff above also would not work if there were more than 2 dimensions.  **The internal attributes that store the coordinates, etc, rely very crucially on order -- dimlabels sets the order, and the corresponding element of, e.g. `axis_coords` is expected to contain the relevant attribute *for that dimension***

pyspecdata/core.py
```diff
@@ -1705,40 +1689,36 @@ def concat(datalist, dimname, chop=False):
...
         # }}}
         # {{{ pull the axis labels from the last item in the list
         if len(datalist[-1].axis_coords) > 0:
             dimlabels = list(datalist[-1].dimlabels)
             axis_coords = list(datalist[-1].axis_coords)
             # print "axis_coords are",axis_coords,"for",dimlabels
-            if dimname in dimlabels:
-                thisindex = dimlabels.index(dimname)
-                dimlabels.pop(thisindex)
-                axis_coords.pop(thisindex)
```
jmfranck: I also removed this

jmfranck: since this won't be executed, given the new code.

jmfranck (changes requested):

pyspecdata/core.py
guinneak198: this looks nice!

pyspecdata/core.py
```diff
@@ -1617,21 +1618,34 @@ def concat(datalist, dimname, chop=False):
...
+        # {{{ check that all the shapes match, too, aside from the dim we concat along
+        shape_check = [
+            list(datalist[j].data.shape) for j in range(len(datalist))
+        ]
+        for j in shape_check:
+            j.pop(dim_idx)
+        assert all(
+            [shape_check[j] == shape_check[0] for j in range(len(shape_check))]
+        ), ("shapes " + str(shape_check) + " are not all equal")
+        # }}}
```
jmfranck, line 1637: reading it again, I realized that we probably wanted more error checks to avoid weird behavior.  Please read these and make sure you're cool with them, and that the example runs the way you wnat

pyspecdata/core.py
```diff
@@ -1617,21 +1618,34 @@ def concat(datalist, dimname, chop=False):
...
+            axis=dim_idx,
+        )
+        concated_ax_coords = np.concatenate(
+            tuple(datalist[j][dimname] for j in range(len(datalist)))
         )
-        concated_ax_coords = np.concatenate(tuple(datalist[j][dimname] for j in range(len(datalist))))
         # }}}
-        retval = datalist[-1].copy(data=False)
-        retval.axis_coords[datalist[-1].axn(dimname)] = concated_ax_coords
+        retval = datalist[0].copy(data=False)
```
jmfranck, line 1647: I started using 0 instead of -1, just because seemed more natural -- no particular reason

jmfranck (approved):

guinneak198 (commented):
These all look good to me and the example still runs as it should

PR #59, opened 2024-09-12
Fix pcolor bug

jmfranck:


.github/workflows/flake8.yml
```diff
@@ -28,8 +28,9 @@ jobs:
           git fetch origin master
           if [ "${{ github.event_name }}" == "pull_request" ]; then
             git fetch origin ${{ github.event.pull_request.head.ref }}
+          else
```
jmfranck, line 31: here I was just fixing a bug with the auto-flake.

examples/pcolor_example.py
```diff
@@ -18,12 +18,44 @@
 
 x = psp.nddata(r_[-5, -2, -1, -0.5, 0, 0.5, 5], "x")
 y = psp.nddata(3 * r_[-5, -2, -1, -0.5, 0, 0.5, 5], "y")
-z = plt.exp(-((y - 2) ** 2) - (x - 0) ** 2 / 2) + 1j * x
+z1 = plt.exp(-((y - 2) ** 2) - (x - 0) ** 2 / 2) + 1j * x
+z2 = 10 * z1
```
jmfranck, line 22: I create a large dataset then a small dataset

examples/pcolor_example.py
```diff
@@ -18,12 +18,44 @@
...
-z.dimlabels
-z.pcolor()
+z1.C.pcolor(scale_independently=True)
+# }}}
+# {{{ plot the larger data
 plt.figure()
+z2.pcolor(scale_independently=True)
+
+
+# }}}
```
jmfranck, line 32: surprisingly, these don't work well.  Notice that I even tried copying the first dataset before running pcolor.

If I comment out either call to pcolor, the other one works correctly.

examples/pcolor_example.py
```diff
@@ -18,12 +18,44 @@
...
+plt.suptitle("colorscales dependent")
+mpb = z1.C.pcolor(ax1=ax_list[0], ax2=ax_list[1])
+z2.C.pcolor(mappable_list=mpb, ax1=ax_list[2], ax2=ax_list[3])
+# }}}
+# {{{ large in first row, then small in second row
+ax_list = new_figure_and_grid()
+plt.suptitle("colorscales dependent")
+mpb = z2.C.pcolor(ax1=ax_list[0], ax2=ax_list[1])
+z1.C.pcolor(mappable_list=mpb, ax1=ax_list[2], ax2=ax_list[3])
+# }}}
```
jmfranck, line 60: here I do both independent scaling, as well as changing which dataset is in which row.  I do that because right now it behaves differently depending on which dataset you call first.

pyspecdata/plot_funcs/pcolormesh.py
```diff
@@ -10,12 +10,15 @@ def pcolormesh(
     ax2=None,
     ax=None,
     scale_independently=False,
+    vmin=None,
+    vmax=None,
```
jmfranck, line 14: these are good to have, but not currently used.

I think that if these are set, you (1) throw an error if `scale_independently` or `force_balanced_cmap` are set, b/c that doesn't make sene and (2) overwrite the values of overall_max and overall_min with these values.

wfkincaid: I don't see why I need to have vmin and vmax overwrite overall_max and _min, I think they should just be whatever they are manually set to and then appended to set_clim for modifying the colorbar

pyspecdata/plot_funcs/pcolormesh.py
```diff
@@ -49,6 +58,10 @@ def pcolormesh(
         under a uniform color scale
     """
     assert len(self.dimlabels) == 2, "currently, this only supports 2D data"
+    if mappable_list is None:
+        # settings the default value to just [] above does NOT create a
+        # new object.  This does
+        mappable_list = list([])
```
wfkincaid, line 64: Should this also be done for vmin_list/vmax_list below?

pyspecdata/plot_funcs/pcolormesh.py
wfkincaid: this way may have worked but its probably better to only make one call to `set_clim` at the end for adjusting the colorbar

PR #60, opened 2024-09-17
Subscripts

guinneak198:
was running into the issue where "t2" didn't match the compile expression because it didn't have dollar signs. I don't want to have to add dollar signs to a dimension every time I want to run cov_mat so I edited


guinneak198:
example (calculate_covariance.py) still works

guinneak198:
To simulate the fail change fieldaxis = "B" and rename the "$B_0$" axis for the example and you will get an error using the master

PR #64, opened 2024-11-04
Colorbar intensity

guinneak198:
so I can have a color bar when only plotting the real (no subplots)

### DIFF
--- a/pyspecdata/plot_funcs/pcolormesh.py
+++ b/pyspecdata/plot_funcs/pcolormesh.py
@@ -107,7 +107,7 @@ def pcolormesh(
             if scale_independently:
                 plt.colorbar(mappable=mappable, ax=thisax)
             elif one_plot:
-                plt.colorbar(mappable=mappable,ax = thisax)
+                plt.colorbar(mappable=mappable, ax=thisax)
             else:
                 pass  # b/c no use for extra colorbar if locked together
         elif j == 1:
--- a/pyspecdata/plot_funcs/pcolormesh.py
+++ b/pyspecdata/plot_funcs/pcolormesh.py
@@ -147,6 +147,7 @@ def pcolormesh(
                     overall_max = -overall_min
             # }}}
         for thismappable in mappable_list:
+            print("okay")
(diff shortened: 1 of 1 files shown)


pyspecdata/plot_funcs/pcolormesh.py
```diff
@@ -84,6 +84,7 @@ def pcolormesh(
         if ax1 is None:
             fig = plt.gcf()
             ax1 = fig.add_subplot(1, 1, 1)
+            one_plot = True
```
jmfranck, line 87: The conditional here is incorrect. You also need to make sure that you are not doing real and imaginary data.

jmfranck: Also, if there is just one plot, then you should set scale independently. There is no need for a different variable.

jmfranck (changes requested):

pyspecdata/plot_funcs/pcolormesh.py
guinneak198: if scale_independently is true then it won't allow setting vmin and vmax for intensity (see lines 131)

guinneak198: also the conditional (if ax1 is None) is in master I can change it but this then won't be minimal changes. On line 74 there is an if for if the data is complex this section is when the data is not complex (which is what I want)

jmfranck:
I realized that this is introducing too much complexity for what you're trying to do.  You can achieve what you want without introducing any new code to pyspecdata:
```
import pyspecdata as psp
import matplotlib.pylab as plt
from numpy import r_


x = psp.nddata(r_[-5, -2, -1, -0.5, 0, 0.5, 5], "x")
y = psp.nddata(3 * r_[-5, -2, -1, -0.5, 0, 0.5, 5], "y")
z1 = plt.exp(-((y - 2) ** 2) - (x - 0) ** 2 / 2)
plt.figure()
plt.suptitle("single plot")
(mpbl,) = z1.pcolor()
plt.colorbar(mpbl)
plt.show()
```

PR #66, opened 2024-12-02
new branch for docstring to avoid conflicts

guinneak198:
detail the docstring


jmfranck (changes requested):
You should refer them to the divide by function, which is preferred.

jmfranck:
> You should refer them to the divide by function, which is preferred. 

Use :func:` to do this (see sphinx docs)

PR #70, opened 2025-01-14
Error tweaks

jmfranck:


pyspecdata/core.py
```diff
@@ -1333,14 +1334,16 @@ def show_array(x, indent=""):
...
-        retval += show_array(self.get_error())
-        if (
-            len(self.dimlabels) > 1
-            or len(self.dimlabels) == 0
-            or self.dimlabels[0] != "INDEX"
-        ):
+        if self.get_error() is not None:
+            retval += "\n\t\t±"
+            retval += show_array(self.get_error())
+        if len(self.dimlabels) > 1 or self.dimlabels[0] != "INDEX":
```
guinneak198, line 1346: If the array is just 1D (for example our list of measured voltages) we don't really care/want dimnames attached with the calcualted error. We just want one value in our returned string.

guinneak198: more a note to myself but playing around with this and a mwe I see that when I make an nddata with the measured voltages (like in verify.py) and run mean with stderr=True the version before would return the mean with attached error as an nddata array and the error was an array and looked really ugly. But with the updates, the code sees it is 1D and so just attaches a p/m and the calcualted error

pyspecdata/core.py
```diff
@@ -3357,7 +3365,10 @@ def mean(self, *args, **kwargs):
                 "return_error kwarg no longer used -- use std kwarg if you"
                 " want to set the error to the std"
             )
-        return_error = process_kwargs([("std", False)], kwargs)
+        return_error, return_stderr = process_kwargs(
+            [("std", False), ("stderr", False)], kwargs
+        )
+        return_error |= return_stderr
```
guinneak198, line 3371: not has option to return std error vs std dev so when we can account for variability across multiple samples (e.g. our V measurements with the scope and AFG)

pyspecdata/core.py
```diff
@@ -3391,6 +3402,8 @@ def mean(self, *args, **kwargs):
                 thiserror = np.std(self.data, axis=thisindex)
                 if np.isscalar(thiserror):
                     thiserror = r_[thiserror]
+                if return_stderr:
+                    thiserror /= np.sqrt(self.data.shape[thisindex])
```
guinneak198, line 3406: see above

pyspecdata/core.py
```diff
@@ -1999,6 +2002,9 @@ def div_units(self, arg1, arg2=None):
         or `d.div_units("s")` to divide the
         data units by seconds.
 
+        In other words, if you pass "a" and the units of your data are
+        in "b", then this returns x, such that (x a)/(b) = 1.
+
```
guinneak198, line 2007: we did this together when reviewing Verify.py and this helps clarify the use of div_units (could help us understand why we multiply vs divide by this return

pyspecdata/core.py
```diff
@@ -1333,14 +1334,16 @@ def show_array(x, indent=""):
             else:
                 return x
 
+        if self.data.size < 2:
+            retval = str(self.data.item())
+            if self.get_error() is not None:
+                retval += "±" + str(self.get_error().item())
+            return retval
```
jmfranck, line 1341: It think address this seperately -- this is not about 1D

pyspecdata/core.py
jmfranck: yes

pyspecdata/core.py
jmfranck: I think it's useful to know:

- what is the `__str__` method? How is it special?
- the block above involves two changes -- one is to wrap lines 1344 and 1345 with an if statment --> what's that for?  The other is to modify the following if statement --> honestly, that's not very important.

pyspecdata/core.py
jmfranck: If I'm understanding correctly, which is a bit off base -- I think that you want to refresh:

- what does `process_kwargs` do?
- what does the `|=` operator do?

and use that to explain what's up here

pyspecdata/core.py
jmfranck: I don't think this is explained properly above, but this one is easiest to just explain here --> if `return_stderr` is set, then it will return the standard error rather than the standard deviation.

In making this comment, I realize that we should probably make `ddof` (see [here](https://numpy.org/doc/stable/reference/generated/numpy.std.html) for explanation) a kwarg of the main function, pass it to std, and default it to 1 when we ask for stderr and 0 when we as for stdev.

pyspecdata/core.py
jmfranck: (the if statement on line 1346 *is* about dimensionality, in contrast the block that ends on 1341)

Also, for this, note that `"INDEX"` is a "default" dimension name -- if we generate a 1D nddata without explicitly naming the dimension then it calls it `"INDEX"` behind the scenes -- e.g. you get this if you do `a = nddata(r_[0:10])`, although the point of this code is to make sure that you would never explicitly see this unless you asked for `a.dimlabels`

pyspecdata/core.py
guinneak198: Using process_kwargs it sees if you passed std or stderr as True. Then based on those it sets return_error or return_stderr to True (whaterver your kwarg was)

guinneak198: Then the |= is an "in place OR operator" so it reassigns return_error to whichever was determined as True in the previous process_kwargs

pyspecdata/core.py
guinneak198: here we removed the if the there are no dimlabels (e.g.` len(self.dimlabels)==0`) so only if there is a dimlabel present do we include the array of errors labeled with the dimlabels

guinneak198: the first if statement we are saying only if there is an error attached should we display the error as an array as part of the return. (It doesn't make sense to have an empty array attached)

guinneak198: the __str__ method I think is so that when we print or want to see the nddata it shows it in the correct layout (e.g. the data points, dimlabels, sizes, and when applicable the error)

guinneak198: ^this is really helpful for taking the full view of the data and associated dimlabels and axes thereof

pyspecdata/core.py
guinneak198: if the self is a single number (e.g. the mean) then there is no need to show the array or dimlabels and axes etc. of the nddata since it's just a number so it just adds the error to the single datapoint and calls it a day for the return

pyspecdata/core.py
guinneak198: should we do that in this PR or you're just commenting for reference?

pyspecdata/core.py
guinneak198: to be specific process_kwargs should be returning the tuple with those set to True or False

pyspecdata/ipy.py
```diff
@@ -161,7 +161,9 @@ def _print_plain_override_for_nddata(arg,p,cycle):
         import IPython.display as d
         arg_copy = arg.copy()
         arg_copy.human_units()
-        if len(arg_copy.dimlabels) == 1:
+        if arg_copy.data.size < 2:
+            print(arg_copy)
+        elif len(arg_copy.dimlabels) == 1:
```
guinneak198, line 166: so that in jupyter notebook we can print the nddata in units even if the data we have is a single value (with or without errors)

pyspecdata/core.py
```diff
@@ -1333,14 +1334,27 @@ def show_array(x, indent=""):
...
+        if self.data.size < 2:
+            val = self.data.item()
+            err = self.get_error()
+            if err is not None:
+                err = err.item()
+                oom_err = int(np.floor(np.log10(err)))  # int takes floor
+                oom_val = int(np.floor(np.log10(val)))  # int takes floor
+                retval = (
+                    "%#0." + str(oom_val - oom_err + 1) + "g ± %#0.2g"
+                ) % (val, err)
```
guinneak198, line 1346: pretty sure this is to get the right sig figs for the value

guinneak198: but also this goes with the comment in the example saying we only use 2 sig figs for the error

examples/error_example.py
```diff
@@ -0,0 +1,72 @@
...
+
+conc_indiv = weight / vol
+conc_indiv.mean(stderr=True)
+print(conc_indiv)
+
+# %%
+# We take the mean, allowing it to accumulate the standard error.
+# Notice how, during string conversion, we always give the standard error
+# 2 significant figures, and then base the significant figures of the
+# number on the error.
```
guinneak198, line 47: should this be repeated? we say this 10 lines up

examples/error_example.py
```diff
@@ -0,0 +1,72 @@
...
+        "Compare one-by-one division,\nfollowed by standard"
+        f" error:\n{conc_indiv}\nto division with error"
+        f" propagation:\n{weight/vol}"
+    ),
+    fontsize=40,
+    ha="center",
+    va="center",
+)
+fig.add_artist(text)
+fig.tight_layout()
```
guinneak198, line 71: personal taste I'd rather this just be printed to the terminal as making an entire figure to just print things seems weird and the figsize is also oddly huge for one sentence. But if you're good with it so am I

pyspecdata/core.py
```diff
@@ -1333,14 +1334,27 @@ def show_array(x, indent=""):
...
+            else:
+                retval = "%#0.5g" % val
+            myunits = self.get_units()
+            if myunits is not None:
+                retval += f" {Q_(myunits).units:~P}"
+            return retval
```
guinneak198, line 1352: ooo fancy I like this so now when we get the returned value with error it has units in the str

pyspecdata/core.py
```diff
@@ -1323,6 +1324,21 @@ def __repr_pretty__(self, p, cycle):
...
+        elif format_spec == "~L":
+            retval = str(self.C.human_units(scale_data=True))
+            retval.replace("±", "\\pm")
+            return re.sub("([^0-9. ]+)", "\;\\\\text{\\1}", retval)
+        else:
+            raise ValueError(
+                "Right now, pyspecdata only responds to the ~P and ~L format"
+                " specs (analogous to pint)"
+            )
+
```
guinneak198, line 1341: format the units as pretty or latex

pyspecdata/core.py
```diff
@@ -1333,14 +1349,27 @@ def show_array(x, indent=""):
...
-        retval += "\n\t\t+/-"
-        retval += show_array(self.get_error())
-        if (
-            len(self.dimlabels) > 1
-            or len(self.dimlabels) == 0
-            or self.dimlabels[0] != "INDEX"
-        ):
+        if self.get_error() is not None:
+            retval += "\n\t\t±"
+            retval += show_array(self.get_error())
```
guinneak198, line 1371: shouldn't be adding the error on if there is no error set

pyspecdata/core.py
```diff
@@ -2052,45 +2096,56 @@ def set_error(self, *args):
...
-                    map(
-                        repr,
-                        [
-                            "Not a valid argument to set_error:",
-                            list(map(type, args)),
-                        ],
+                raise TypeError(
+                    "Passed one argument, but I don't understand its type: "
+                    + str(type(args[0]))
+                )
```
guinneak198, line 2126: when the error of the data is one value

pyspecdata/core.py
```diff
@@ -2052,45 +2096,56 @@ def set_error(self, *args):
...
-        if (len(args) == 1) and np.isscalar(args[0]):
-            if args[0] == 0:
-                args = (np.zeros_like(self.data),)
+        if len(args) == 1:
+            if np.isscalar(args[0]):
+                logger.debug("converting scalar")
+                if args[0] == 0:
+                    args = (np.zeros_like(self.data),)
+                else:
+                    args = (np.ones_like(self.data) * args[0],)
```
guinneak198, line 2105: if the error is scalar, convert to an array

pyspecdata/core.py
```diff
@@ -2052,45 +2096,56 @@ def set_error(self, *args):
...
+                + str(type(args[0]))
+                + "is numpy?: "
+                + str(isinstance(args[0], np.ndarray))
+            )
+            if isinstance(args[0], np.ndarray):
+                self.data_error = np.reshape(args[0], np.shape(self.data))
+            elif isinstance(args[0], list):
+                self.data_error = np.reshape(
+                    np.array(args[0]), np.shape(self.data)
+                )
```
guinneak198, line 2117: make sure the shape of the error array is correct according to the shape of the data

pyspecdata/core.py
```diff
@@ -2052,45 +2096,56 @@ def set_error(self, *args):
...
+            elif args[0] is None:
+                self.data_error = None
+            elif np.isscalar(args[0]):
+                self.data_error = args[0]
             else:
```
guinneak198, line 2122: this doesn't make sense didn't we already have an if the args[0] is scalar above where we convert it to array?

pyspecdata/core.py
```diff
@@ -2052,45 +2096,56 @@ def set_error(self, *args):
...
+                elif np.isscalar(args[1]):
+                    self.axis_coords_error[self.axn(args[0])] = args[
+                        1
+                    ] * np.ones_like(self.getaxis(args[0]))
+                else:
+                    raise TypeError(
+                        "you passed two arguments, starting with a string."
+                        " But, this was not followed by numpy array or scalar,"
+                        " but rather "
+                        + str(type(args[1]))
```
guinneak198, line 2140: I'm thinking his is so you can set the error of a specific axis to an array of errors

pyspecdata/core.py
```diff
@@ -2500,6 +2555,15 @@ def __mul__(self, arg):
...
         retval.set_error(Rerr)
+        unitA = Q_(
+            "dimensionless" if self.get_units() is None else self.get_units()
+        )
+        unitB = Q_(
+            "dimensionless" if arg.get_units() is None else arg.get_units()
+        )
+        unit_ret = f"{(unitA*unitB).units:~P}"
+        unit_ret = None if len(unit_ret) == 0 else unit_ret
+        retval.set_units(unit_ret)
```
guinneak198, line 2566: to handle unit conversion if the error is in different units from the data or axis

pyspecdata/core.py
```diff
@@ -2614,6 +2678,15 @@ def __div__(self, arg):
...
         retval.set_error(Rerr)
+        unitA = Q_(
+            "dimensionless" if self.get_units() is None else self.get_units()
+        )
+        unitB = Q_(
+            "dimensionless" if arg.get_units() is None else arg.get_units()
+        )
+        unit_ret = f"{(unitA/unitB).units:~P}"
+        unit_ret = None if len(unit_ret) == 0 else unit_ret
+        retval.set_units(unit_ret)
```
guinneak198, line 2689: unit conversion for division

pyspecdata/core.py
```diff
@@ -592,9 +594,9 @@ def plot(*args, **kwargs):
         myy = args[1]
     if len(args) == 3:
         myformat = args[2]
-    if np.isscalar(myx):
+    if scalar_or_zero_order(myx):
         myx = np.array([myx])
-    if np.isscalar(myy):
+    if scalar_or_zero_order(myy):
```
guinneak198, line 599: replaces the if scalar or if its an ndarray with a shape of 0

pyspecdata/core.py
```diff
@@ -2052,45 +2094,56 @@ def set_error(self, *args):
...
+                    self.axis_coords_error[self.axn(args[0])] = args[1]
+                elif scalar_or_zero_order(args[1]):
+                    self.axis_coords_error[self.axn(args[0])] = args[
+                        1
+                    ] * np.ones_like(self.getaxis(args[0]))
+                else:
+                    raise TypeError(
+                        "you passed two arguments, starting with a string."
+                        " But, this was not followed by numpy array or scalar,"
+                        " but rather "
```
guinneak198, line 2137: can set the error of specific axes

PR #71, opened 2025-01-21
Postproc none

guinneak198:


pyspecdata/__init__.py
```diff
@@ -15,32 +15,32 @@
...
-    whereblocks,
-    box_muller,
     dp,
+    emptytest,
     fa,
+    fname_makenice,
+    init_logging,
+    inside_sphinx,
+    level_str_to_int,
+    lsafe,
```
jmfranck, line 31: here I'm just sorting the function names.

pyspecdata/load_files/__init__.py
```diff
@@ -221,6 +221,19 @@ def look_inside(inp_directory):
...
+def register_proc_lookup(newdict):
+    """this updates the dictionary at
+    pyspecdata.load_files.postproc_lookup
+    this is equivalent to passing newdict to the lookup
+    keyword argument of `find_file`.
+
+    Remember that you only have to pass the keyword argument or call this
+    function once per script, and the effects are persistent!
+    """
+    postproc_lookup.update(newdict)
```
guinneak198, line 233: here he's making the update a function whereas before it was just in the code always called

pyspecdata/load_files/__init__.py
```diff
@@ -324,10 +337,17 @@ def find_file(
         available in pySpecData, but also, see the `lookup` argument,
         below).
 
+        Note that we call this "postprocessing" here because it follows
+        the data organization, *etc.*, performed by the rest of the file
+        in other contexts, however, we might call this "preprocessing"
+
```
guinneak198, line 343: good explanation since we do use the terms interchangeably

pyspecdata/load_files/__init__.py
```diff
@@ -221,6 +221,19 @@ def look_inside(inp_directory):
...
+    this is equivalent to passing newdict to the lookup
+    keyword argument of `find_file`.
+
+    Remember that you only have to pass the keyword argument or call this
+    function once per script, and the effects are persistent!
+    """
+    postproc_lookup.update(newdict)
+    return
+
+
```
guinneak198, line 236: I think this is doing what you explain `[here](https://jmfrancklab.slack.com/archives/C06LNFNP2HM/p1737499687080149?thread_ts=1737467514.307219&cid=C06LNFNP2HM)`  but you made it a function so we can call whenever

PR #73, opened 2025-01-22
indent error was giving wrong ph2 labels

guinneak198:
dumb indent error was causing the ph2 dims to be 0.0,1.0,2.0 but fixed


jmfranck:
conflicts with recent changes, so I'm closing, but going to push a fix straight to master

PR #74, opened 2025-01-27
edit cov_mat so it works properly

guinneak198:
see https://jmfrancklab.slack.com/archives/C0811PEGU5B/p1738009394702829

### DIFF
--- a/examples/covariance.py
+++ b/examples/covariance.py
@@ -1,0 +1,23 @@
+import pyspecdata as psd
+import pyspecProcScripts as psdpr
+
+data = psd.find_file(
+    "241003_27mM_TEMPOL_amp0p1_var_tau_pm_echo",
+    exp_type="ODNP_NMR_comp/Echoes",
+    expno="echo_1",
+    lookup=psdpr.lookup_table,
+)
+data.ift("t2")
+# FID slice
+data = data["t2":(0, None)]
+data *= 2
+data["t2":0] *= 0.5
+# }}}
+# relabel nScans axis with indices
+data.setaxis("nScans", "#")
+with psd.figlist_var() as fl:
+    covariance_matrix = psdpr.select_pathway(
+        data, data.get_prop("coherence_pathway")
+    ).cov_mat("nScans")
+    fl.next("2D Covariance Matrix")
+    fl.image(abs(covariance_matrix))
--- a/examples/covariance.py
+++ b/examples/covariance.py
@@ -16,7 +16,6 @@ data["t2":0] *= 0.5
 # relabel nScans axis with indices
 data.setaxis("nScans", "#")
 with psd.figlist_var() as fl:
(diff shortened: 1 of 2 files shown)


pyspecdata/core.py
```diff
@@ -55,7 +55,6 @@
 from .general_functions import lsafe as orig_lsafe
 from .general_functions import (
     Q_,
-    ureg,
```
jmfranck, line 58: Why??

pyspecdata/core.py
```diff
@@ -3544,22 +3543,23 @@ def cov_mat(self, along_dim):
         Parameters
         ==========
         along_dim:  str
-            the "observations" dimension of the data set (as opposed to
-            the variable)
+            the "variables" dimension of the data set (as opposed to
+            the variable). Note when rowvar = True for np.cov the rows are the
+            variables.
```
jmfranck, line 3548: I disagree with this change. You should specify the dimension that represents repeated measurements

examples/covariance.py
```diff
@@ -0,0 +1,23 @@
...
+data["t2":0] *= 0.5
+# }}}
+# relabel nScans axis with indices
+data.setaxis("nScans", "#")
+with psd.figlist_var() as fl:
+    covariance_matrix = psdpr.select_pathway(
+        data, data.get_prop("coherence_pathway")
+    ).cov_mat("nScans")
+    fl.next("2D Covariance Matrix")
+    fl.image(abs(covariance_matrix))
```
jmfranck, line 23: I agree very much with making an example that shows the covariance of an FID since we are pretty sure we know what it should look like.  (As opposed to ESR). If it were only for the changes that you made to the function, then I would close the pull request, but I like this. 

Please rename the function so that it's clear we're talking about NMR here, and so that it is named more like the other examples.

pyspecdata/core.py
jmfranck: The reason why that makes sense is that that is the same dimension along which you would take the standard deviation, and the square of the standard deviation should match the diagonal of the covariance matrix.

In fact, in your example, you should explicitly show that that is true. That's where you're getting caught up.

jmfranck (changes requested):

pyspecdata/core.py
guinneak198: flake said it wasn't used

pyspecdata/core.py
guinneak198: please look at the documentation for numpy cov. the rows (e.g. nScans) are termed the variables and the columns (e.g. t2) are the observations. I think it makes more sense to keep the phrasing of the function we are using

guinneak198: https://numpy.org/doc/2.1/reference/generated/numpy.cov.html

PR #75, opened 2025-01-30
Cov example

guinneak198:
example to demonstrate covariance code

### DIFF
--- a/examples/proc_NMR_covariance.py
+++ b/examples/proc_NMR_covariance.py
@@ -1,49 +1,81 @@
+""" Calculate covariance matrix for NMR signal
+==============================================
+First, we show (both with pyspec and then with raw nddata) that the early
+points of the FID change *together* due to the fact that the peak is moving
+around in frequency space. 
+This means that the phase of the time domain points is not fixed -- as the
+phase varies, the real and imaginary components of the signal *change together*
+- covariance.
+However, we show that with the application of a correlation alignment run prior
+to calculating the covariance, this effect and therefore covariance is
+decreased.
+"""
+
 import pyspecdata as psd
 import numpy as np
+import pyspecProcScripts as psdpr
+import matplotlib.pyplot as plt
 
+signal_pathway = {"ph1": 1}
 d = psd.find_file(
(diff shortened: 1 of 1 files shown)


jmfranck:
this probably belongs on proc_scripts.  But also, it fails with an error about shifting the ft when I use the current master on both proc scripts and pyspecdata

guinneak198:
Shouldn't it be a pyspecdata example since it is displaying the uses etc of the cov_mat function which is a pyspecdata function?

examples/covariance_example.py
```diff
@@ -0,0 +1,31 @@
...
+    expno="echo_1",
+    postproc="none",
+).squeeze()
+with psd.figlist_var() as fl:
+    thisdata = thisdata["ph1", 0]  # pull only one phcyc step to simplify
+    # I explictly reorder so I know which dimension is which when I do raw
+    # numpy and matplotlib
+    thisdata.reorder(["t2", "nScans"])
+    mycov = abs(np.cov(thisdata.data))
+    fl.next("covariance")
```
jmfranck, line 23: label as numpy

jmfranck: if the following is ABS, this one should be too

examples/covariance_example.py
```diff
@@ -0,0 +1,31 @@
...
+    fl.image(mycov)
+    fl.next("covariance - pyspecdata")
+    fl.image(abs(thisdata.C.cov_mat("nScans")))
```
jmfranck, line 26: why are you making a copy here?

examples/covariance_example.py
guinneak198: it's already abs (see mycov = )

examples/covariance_example.py
guinneak198: added comment. I make copy since this function renames and resets axes from t2 to t2_i and t2_j.

guinneak198: ^ thus the slicing of the FID tail becomes error prone saying t2 is no longer the axis label etc.

PR #76, opened 2025-01-31
Fix register axis

jmfranck:


pyspecdata/fourier/ft_shift.py
```diff
@@ -229,6 +229,7 @@ def ft_new_startpoint(self, axis, which_domain, value=None, nearest=False):
     elif which_domain == "f":
         which_domain = "freq"
         thisconditional = True
+    other_domain = tuple(set(("freq","time"))-set((which_domain,)))[0]
```
guinneak198, line 232: allows us to ft a dimension other than frequency or time for example field

pyspecdata/fourier/ft_shift.py
```diff
@@ -243,10 +244,8 @@ def ft_new_startpoint(self, axis, which_domain, value=None, nearest=False):
             )
     if value is not None:
         n_du = (orig_u - value) / du  # number of du's shifted by
-        if abs((n_du - round(n_du)) / n_du) > 1e-3:
-            if nearest:
-                value = round((value - orig_u) / du) * du + orig_u
-            else:
+        if not n_du==0 and abs((n_du - round(n_du)) / n_du) > 1e-3:
+            if nearest is None:
```
guinneak198, line 248: to avoid dividing by 0

pyspecdata/fourier/ft_shift.py
```diff
@@ -266,9 +265,14 @@ def ft_new_startpoint(self, axis, which_domain, value=None, nearest=False):
...
-        else:
-            value = orig_u - round(n_du) * du
+            if nearest:
+                value = round((value - orig_u) / du) * du + orig_u
+            else:
+                value = orig_u - round(n_du) * du
     self.set_ft_prop(axis, ["start", which_domain], value)
+    self.set_ft_prop(axis, [which_domain, "not", "aliased"], None)
+    if nearest is False:
+        self.set_ft_prop(axis, [other_domain, "not", "aliased"], True)
```
guinneak198, line 275: using ft_new_startpoint to set FT property (has it been FTed with the new startpoint defined?)

pyspecdata/fourier/ft_shift.py
jmfranck: That's not the point of this. We refer to the domains as frequency and time whether they are or are not

pyspecdata/fourier/ft_shift.py
jmfranck: It's actually unsetting

jmfranck: Add your missing the second part

pyspecdata/fourier/ft_shift.py
guinneak198: ah so this is more compact to define the other domain vs putting it inside each if or elif independently (for example inside the if `which_doman == "t": other_domain = "f"`)

pyspecdata/fourier/ft_shift.py
guinneak198: It's unsetting the ft property unless nearest is False in which case it is setting it to True

examples/covariance_example.py
```diff
@@ -0,0 +1,31 @@
...
+    mycov = abs(np.cov(thisdata.data))
+    fl.next("covariance")
+    fl.image(mycov)
+    fl.next("covariance - pyspecdata")
+    fl.image(abs(thisdata.C.cov_mat("nScans")))
+    # now I zoom in to show that when I'm in the part of the FID that's
+    # just noise (no shifting frequencies, I get the diagonal covariance
+    # that I expect)
+    fl.next("covariance - FID tail")
+    fl.image(abs(thisdata["t2":(0.6,0.7)].cov_mat("nScans")))
```
jmfranck, line 31: why is this here?

pyspecdata/axis_manipulation/register_axis.py
```diff
@@ -1,36 +1,41 @@
...
         idx = np.argmin(abs(x - v))
-        offset = (v # where I want to be
-                - x[idx]) # where I actually am
-        offset += x[0] # since the following takes the startpoint
+        offset = v - x[idx]  # where I want to be  # where I actually am
+        offset += x[0]  # since the following takes the startpoint
         if self.get_ft_prop(k):
-            self.ift(k).ft_clear_startpoints(k,t='current',f=offset,nearest=False)
-            self.set_ft_prop(k,'freq_not_aliased').ft(k)
+            self.ift(k).ft_new_startpoint(k, "f", value=offset, nearest=False)
```
jmfranck, line 31: github comment to explain what this is doing?

examples/covariance_example.py
guinneak198: I'm confused on your question... the comment above says we zoom in to show the noise has the diagonal covariance that's expected

pyspecdata/axis_manipulation/register_axis.py
guinneak198: resets the startpoint of the axis to the offset (calculated above) and then sets the ft prop to say the frequency axis is NOT aliased (since we just reset the startpoint

pyspecdata/fourier/ft_shift.py
guinneak198: above we set this property based on the startpoints we set either in the f domain or t domain

examples/covariance_example.py
guinneak198: oh I see this example shouldn't be part of this PR somehow got pulled in or something?

pyspecdata/fourier/ft_shift.py
guinneak198: this business with nearest is whether we have shifted by the nearest integral number of datapoints vs. the alternative which results in a phase dependent shift

PR #81, opened 2025-04-04
Fix dcct

jmfranck:


guinneak198:
looks good to me and I tested with the correlation alignment example that has an extra Delta p1 dimension and it works

PR #135, opened 2026-06-29
Zenodo tweak

jmfranck:


jmfranck:
I'm OK with this once it passes all the tests.

jmfranck:
note that I'm imlementing the single-use stuff in the ciqtek branch

PR #136, opened 2026-06-30
Fixing lmfitdata weight issue

atahangarip:


pyspecdata/lmfitdata.py
```diff
@@ -393,10 +393,7 @@ def fit(self, use_jacobian=True):
                 self.guess_parameters,
                 fcn_args=(sigma,),
             )
-        if use_jacobian:
-            out = themin.leastsq(Dfun=self.jacobian, col_deriv=True)
-        else:
-            out = themin.leastsq()
+        out = themin.leastsq(Dfun=self.jacobian, col_deriv=True)
```
jmfranck, line 402: you want to preserve the ability not to use the jacobian.  Sometimes, it's just poorly behaved.

PR #139, opened 2026-07-10
fix jacobian with transform

jmfranck:
- **codex turn with review**
- **tweak and flake happy**


jmfranck:
this complains it can't import Q_ from pyspecdata.  This is likely because the import logic got messed up.  It should absolutely be able to import Q_ from pyspecdata.

jmfranck:
the test packages import stuff differently -- this is probably arising from that

